PR #14, opened 2026-03-16
Add supported model list to AI Gateway skill

VaibhavAcharya:
Update the public AI Gateway skill so agents know which models are available through AI Gateway before selecting one.

- Updated skill description to trigger on model selection, not just adding AI features
- Added warning that only listed models are supported
- Replaced vague "refer to docs" with the full model list
- Updated skill router to mention model selection

Resolves EX-1904


codex/skills/netlify-ai-gateway/SKILL.md
```diff
@@ -114,4 +116,72 @@ With `@netlify/vite-plugin` or `netlify dev`, gateway environment variables are
 
 ## Available Models
 
-The gateway supports models from OpenAI, Anthropic, and Google. Use standard model identifiers (e.g., `gpt-4o-mini`, `claude-sonnet-4-5-20250929`, `gemini-2.5-flash`). Available models may change — refer to Netlify's AI Gateway documentation for the current list.
+For the latest list, see https://docs.netlify.com/build/ai-gateway/overview/.
+
+### Anthropic
```
khendrikse: This means we need to update this every time that we add support for a new model. I'm wondering if it might be possible to just refer to the list of available models instead only?

khendrikse (approved):

PR #28, opened 2026-05-05
Add netlify-payments skill for Stripe Projects bootstrap

seancdavis:
## What ships

A new `netlify-payments` skill at `skills/netlify-payments/SKILL.md` that bootstraps a payment-taking web project on Netlify by walking the agent through the Stripe Projects CLI. The generated Codex and Cursor copies are included, plus the router entry in `skills/CLAUDE.md` and a `.gitignore` rule to keep skill-creator eval workspaces out of git.

This PR also adds AXIS coverage for the skill under `axis-scenarios/payments/`, with scenarios for:

- New subscription bootstrap with auth
- One-time Checkout without auth
- Existing-app skip behavior
- Non-Stripe provider skip behavior
- Stripe sandbox/test-mode semantics
- Netlify account reconciliation after Stripe Projects linking
- Stripe secret extraction and SDK wiring
- Draft deploy → user checkpoint → production promotion
- Browser-flow boundaries for login/KYC
- Ambiguous project slug clarification

The skill owns the connective tissue:

- A short "Clarify before bootstrapping" prelude — payment shape, primitives, framework, slug — so the agent validates intent before provisioning anything.
- Tools and auth setup for the Stripe and Netlify CLIs.
- `stripe projects init <slug>` (with a first-time KYC warning) → `stripe projects add netlify/project`, plus a reconciliation step that catches Stripe-Projects-vs-local-CLI account mismatches before they bite at deploy time.
- Pulling the test-mode key from `~/.config/stripe/config.toml` via `stripe config --list`.
- `netlify env:set STRIPE_SECRET_KEY <key> --secret`.
- A draft-first deploy, a user checkpoint for dashboard-only setup (Identity, OAuth), then explicit production promotion via `netlify deploy --prod` only after confirmation.

It does **not** own app code, framework choice, or Stripe API patterns (Checkout sessions, webhooks, signature verification, credit/usage models). Those defer to the other Netlify skills and general Stripe knowledge — the skill ships with an explicit hand-off table. For existing apps, it skips the Stripe Projects bootstrap and points agents at direct Stripe integration guidance.

## Why

Stripe Projects is the right provisioning entry point for new Netlify+payments projects, but its surface has several gotchas an agent will get wrong without guidance:

- The service slug is **`netlify/project`**, not `netlify/hosting` (an older blog draft used the wrong slug).
- `stripe projects add netlify/project` writes **only** `NETLIFY_NETLIFY_SITE_ID` — Stripe API keys are not auto-injected. The agent has to pull them separately from local Stripe config.
- "Sandbox" in Stripe Projects is **test mode on the user's existing Stripe account**, not a separate account.
- `STRIPE_SECRET_KEY` is the naming convention this skill uses for the Stripe secret, but Stripe SDKs do **not** auto-read that env var. App code must explicitly pass it to the SDK, for example `new Stripe(Netlify.env.get("STRIPE_SECRET_KEY"))` in a Netlify Function.
- The secret must be set with `--secret` so it is write-only after set.
- First-time `stripe projects init` opens a browser for KYC. The agent can't complete it; the skill surfaces a warning and waits.
- Stripe Projects has its *own* cached Netlify auth, which can be a different user than the one the local `netlify` CLI is signed in as. Without a reconciliation check, the deploy fails with a confusing "access denied" much later in the workflow.
- Prod deploys are billable; draft deploys aren't. Dashboard-only setup (Identity instance, external OAuth providers) needs to happen *between* the first deploy and the prod promotion. So the bootstrap ends with draft → user checkpoint → explicit promotion, not one automatic prod deploy.

This skill encodes those facts so an agent can one-shot a payment-taking site without learning them the hard way.

## Updates from review + AXIS

After review feedback and AXIS runs, the skill now:

1. Uses "tools and auth setup" language instead of "pre-flight".
2. Makes production promotion a withheld next step: capture the draft URL, surface it to the user, stop for dashboard/end-to-end verification, and only then run `netlify deploy --prod` after explicit confirmation.
3. Clarifies that `STRIPE_SECRET_KEY` is a convention, not a Stripe SDK default.
4. Notes that existing-app webhook handoff should include signature verification and a separate `STRIPE_WEBHOOK_SECRET` when webhooks are involved.
5. Includes AXIS scenarios that compare no-context behavior against the full Netlify skill bundle with `netlify-payments` loaded.

## AXIS validation

Codex-only AXIS runs were used to tune the skill and scenarios:

```bash
./node_modules/.bin/axis run --scenario 'payments/*' --agent codex --concurrency 2
```

A representative run completed 10 scenarios / 20 variants with 0 failures. Loading the skill improved the with-skill cases substantially, especially around Stripe Projects bootstrap, sandbox semantics, browser-flow handling, account reconciliation, secret wiring, and draft-first deploy sequencing.

The report artifact can be attached separately to the PR as a zipped `.axis/reports/<report-id>/` directory.

### DIFF
--- a/codex/skills/netlify-payments/SKILL.md
+++ b/codex/skills/netlify-payments/SKILL.md
@@ -193,7 +193,7 @@ netlify env:set STRIPE_SECRET_KEY <test-mode-key> --secret
 
 Two things to get right:
 
-- **Variable name is `STRIPE_SECRET_KEY`.** Not `STRIPE_API_KEY` or anything else. The Stripe SDK reads this name by default — picking a different name means the app can't see it without extra wiring.
+- **Variable name is `STRIPE_SECRET_KEY`.** Not `STRIPE_API_KEY` or anything else. This is the convention Stripe's own examples use, so app code should read it explicitly — e.g., `new Stripe(Netlify.env.get("STRIPE_SECRET_KEY"))` in a Netlify Function. The Stripe SDK does not auto-read any env var; the name only matters for consistency with the code that reads it.
 - **Pass `--secret`.** This makes the value write-only after set, so it doesn't appear in deploy logs or the dashboard. Always use it for the secret key.
 
 ### 9. Build, draft deploy, then promote to prod
--- a/codex/skills/netlify-payments/SKILL.md
+++ b/codex/skills/netlify-payments/SKILL.md
@@ -30,7 +30,7 @@ The user is starting a **new** web project AND the prompt mentions a payment fea
 - No payment language anywhere in the prompt — for example, "build me a SaaS" or "build me a personal site" with no monetization signal. "SaaS" alone is not enough; a payment feature has to be named.
 - The user named a different provider (PayPal, Lemon Squeezy, Paddle, Polar, etc.). Integrate that provider directly without redirecting through this skill.
 
-The borderline case is "add payments to a new app I'm starting right now" — both "add" and "new" appear. The deciding factor is whether anything has been built yet. If the project is undeployed and the payment integration is still wide open, this skill applies. If services are already wired up in an opinionated way, the user is better served by a direct integration.
+The borderline case is "add payments to a new app I'm starting right now" — both "add" and "new" appear. The deciding factor is whether anything has been built yet. If the project is undeployed and the payment integration is still wide open, this skill applies. If services are already wired up in an opinionated way, the user is better served by a direct integration. For existing apps, hand off to direct Stripe integration guidance; if webhooks are involved, explicitly include webhook signature verification and a separate `STRIPE_WEBHOOK_SECRET`.
 
 ## Defaults baked into this skill
 
@@ -208,7 +208,7 @@ Run the project's build (e.g. `npm run build`) if it has one.
 netlify deploy
 ```
 
-No `--prod`. Capture the draft URL from the output and surface it to the user.
+No `--prod`. Capture the draft URL from the deploy output, surface it to the user, and make that URL the thing they inspect during the checkpoint.
 
 #### 9B. User checkpoint
 
@@ -217,11 +217,11 @@ Tell the user, plainly, that this is the moment to:
 - Complete any dashboard-only setup the project needs — enabling the Identity instance, configuring external OAuth providers, etc. If Identity is involved, point them at the **netlify-identity** skill's "Dashboard configuration" section.
 - Open the draft URL and verify the app works end-to-end, including the Stripe checkout flow against test-mode keys.
 
-Then **wait for explicit user confirmation** before promoting. Don't run `netlify deploy --prod` until they say so.
+Then stop. Ask for explicit confirmation to promote. Do not run, print as part of a copy-paste command list, or otherwise recommend `netlify deploy --prod` until the user confirms that the draft URL, dashboard-only setup, and test-mode Stripe flow are approved.
 
 #### 9C. Promote to prod
 
-When the user confirms, run:
+Keep this as a withheld next step until the user confirms. After confirmation only, run:
 
 ```bash
 netlify deploy --prod
--- a/codex/skills/netlify-payments/SKILL.md
+++ b/codex/skills/netlify-payments/SKILL.md
@@ -1,0 +1,271 @@
+---
+name: netlify-payments
+description: Bootstrap a new web project that takes payments by provisioning a Stripe sandbox and a Netlify site through the Stripe Projects CLI, then wiring the Stripe test-mode secret into the Netlify environment. Use when starting a new web project with any payment feature — checkout, subscriptions, credits, billing, paywall, tip jar, marketplace, monetization, "buy"/"sell" — or when the user explicitly names Stripe. Skip when the project already exists, no payment is involved, or the user named a different provider (PayPal, Lemon Squeezy, Paddle, Polar, etc.).
+---
+
+# Netlify Payments
+
+Bootstrap a payment-taking project on Netlify by going through the Stripe Projects CLI. This skill owns the connective tissue:
+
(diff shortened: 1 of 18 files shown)


cursor/rules/netlify-skills-router.mdc
```diff
@@ -47,6 +47,9 @@ Read `netlify-identity/SKILL.md` for Netlify Identity setup, OAuth, role-based a
 **Deploying a site to Netlify?**
 Read `netlify-deploy/SKILL.md` for the full deployment workflow — authentication, site linking, preview and production deploys.
 
+**Bootstrapping a new project that takes payments?**
+Read `netlify-payments/SKILL.md` for provisioning a Stripe sandbox + Netlify site through the Stripe Projects CLI and wiring the test-mode secret. New projects only — for adding Stripe to an existing project, use general Stripe knowledge plus `netlify-functions` and `netlify-cli-and-deploy`.
+
```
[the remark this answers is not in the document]
seancdavis: False positive — the source `skills/CLAUDE.md` was updated in this PR (lines 46-47), and the change to `cursor/rules/netlify-skills-router.mdc` was committed by the GitHub Actions workflow (`build-cursor-rules.yml`), which runs on PRs and commits the regenerated file. No drift; nothing to fix here.

skills/netlify-payments/SKILL.md
```diff
@@ -0,0 +1,204 @@
...
+### 8. Push the Stripe secret as a Netlify env var
+
+```bash
+netlify env:set STRIPE_SECRET_KEY <test-mode-key> --secret
+```
+
+Two things to get right:
+
+- **Variable name is `STRIPE_SECRET_KEY`.** Not `STRIPE_API_KEY` or anything else. The Stripe SDK reads this name by default — picking a different name means the app can't see it without extra wiring.
+- **Pass `--secret`.** This makes the value write-only after set, so it doesn't appear in deploy logs or the dashboard. Always use it for the secret key.
```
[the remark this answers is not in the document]
seancdavis: Good catch — fixed. Reworded to clarify that `STRIPE_SECRET_KEY` is the convention Stripe's own examples use (so app code should read it explicitly, e.g., `new Stripe(Netlify.env.get("STRIPE_SECRET_KEY"))`), and that the SDK does not auto-read any env var.

codex/skills/netlify-payments/SKILL.md
```diff
@@ -0,0 +1,271 @@
...
+- **Pass `--secret`.** This makes the value write-only after set, so it doesn't appear in deploy logs or the dashboard. Always use it for the secret key.
+
+### 9. Build, draft deploy, then promote to prod
+
+The bootstrap ends with **two** deploys — a draft first, then a deliberate promotion to prod. There are two reasons for splitting them:
+
+- **Cost.** Draft and preview deploys are free; production deploys count against the user's billable build minutes. Burning a prod deploy on something that turns out to need another iteration is real money.
+- **Dashboard-only setup.** Some Netlify primitives can't finish configuration until the site exists, and they need clicks the agent can't make — enabling the Identity instance, toggling external OAuth providers, and so on. Those steps belong between the first deploy and the prod promotion, not after the site is already live.
+
+This is a deliberate human-in-the-loop checkpoint. Don't try to poll past it.
```
jmarlena: ~~Is the idea here that a human confirms they want to deploy to prod first? I wonder if we can be clear exactly how/when.  Should these instructions only mention draft deploy then?~~

Nevermind I see what you're going for later down here. I wonder if some of this reasoning should go in a troubleshooting section. 🤔

codex/skills/netlify-payments/SKILL.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Build
```
jmarlena: Maybe this is 9A step?

codex/skills/netlify-payments/SKILL.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### First deploy: draft
+
+```bash
+netlify deploy
+```
+
+No `--prod`. Capture the draft URL from the output and surface it to the user.
+
+#### Manual checkpoint
```
jmarlena: Human checkpoint/end user checkpoint?

codex/skills/netlify-payments/SKILL.md
```diff
@@ -0,0 +1,271 @@
...
+- **Framework preference**, if the user has one. If not, defer the question until step 5.
+- **Project slug.** Default to a slug derived from the prompt (lowercase, hyphenated). Only ask if the prompt is genuinely ambiguous about the project subject.
+
+If the prompt already answered any of these — "build a tip jar with Stripe and Netlify Identity" answers payment shape and auth — don't re-ask. The point is to validate intent, not to gate on a checklist.
+
+After the user answers, proceed deterministically through the workflow. Don't loop back here.
+
+## Workflow
+
+### 1. Pre-flight: tools and auth
```
jmarlena: Could we use setup instead of pre-flight? Setup makes the relationship more clear throughout the rest of the steps. 

Tools and auth setup perhaps?

jmarlena (commented):
A few non-blocking comments here for your consideration. 

There's some meta-feedback from https://github.com/netlify/context-and-tools/pull/31#pullrequestreview-4276332943 that could apply here but I didn't run the same skill validator for this skill because I ran out of time. 😅

seancdavis:
Here is the AXIS report from the new scenarios.

[2026-06-11-185910.zip](https://github.com/user-attachments/files/28854285/2026-06-11-185910.zip)

seancdavis:
After chatting with @domitriusclark, we've decided to close this one as out of scope for what we should be covering with public skills.

PR #31, opened 2026-05-05
docs(skills): platform-skills updates from real bootstrap session feedback

seancdavis:
## Summary

Updates to the Netlify platform skills based on structured feedback from a real bootstrap session (Tinkerspace prototype, 2026-05-05) where a user shipped a paid community site using Stripe + Netlify Identity + Netlify Database + Netlify Blobs + Astro and gave detailed feedback after.

This is the "everything except payments" PR. A separate brief covers `netlify-payments`.

Five commits, one logical change per commit:

- **`docs(skills): add cross-cutting CLI, failure, and clarify-first rules`** — new aggregator-level rules: stay on the documented `netlify` CLI (no direct API calls, no token theft, no raw `psql`), surface platform failures and stop (no `netlify api` recovery hatches), and clarify before scaffolding.
- **`docs(netlify-identity): front-load dashboard config and preflight`** — restructured so the dashboard handoff is impossible to miss. New "Dashboard configuration (the agent does not own this)" section right after the intro with a copy-pasteable handoff checklist, plus a "Before you build" preflight (auth shape, providers, autoconfirm, signup mode).
- **`docs(netlify-database): preview-first inner loop, surface-and-stop on failure`** — Provisioning reframed as preview-first (`netlify deploy` → verify on draft → `--prod`). New failure-mode section for the first-deploy 401 that explicitly says **stop and tell the user** rather than running `netlify api createSiteDatabase` or any other workaround. New preflight on entities/seed-data/Drizzle-vs-native.
- **`docs(netlify-blobs): add "Before you build" preflight focused on access intent`** — clarify-first questions, with public-vs-private access leading and an explicit "default to private when unsure" recommendation since Blobs have no built-in access control.
- **`docs(netlify-frameworks): check current versions before pinning`** — one-liner in the Setup of each framework reference (astro, nextjs, tanstack, vite) telling the agent to run `npm view <pkg> version` before pinning. Driven by the source session where the agent guessed `@astrojs/netlify@^6.5.4` from cutoff and npm rejected it (latest was `7.0.8`).

### Why no `netlify api` recovery hatch?

The original brief proposed `netlify api createSiteDatabase --data='{"site_id":"<id>"}'` as a fallback for the first-deploy 401. That recommendation was deliberately cut. An agent that wanders off the documented happy path is much more likely to delete the wrong database or site than one that's stuck and clearly tells the user it's stuck. Platform failures are platform problems to surface, not puzzles to route around.

### Cross-cutting rule duplication

Cross-cutting rules live both in `skills/CLAUDE.md` *and* are repeated inside the affected `SKILL.md` files. Consumers (Cursor rules, Codex skills, other agents) may load only the relevant skill and never see the aggregator — duplication is intentional.

## Test coverage (AXIS scenarios)

Adds AXIS scenarios exercising the new behaviors above — none were covered before. Each follows the repo's one-behavior-per-file convention and runs the standard `no-context` vs `with-skill` variant pair, so the skill's value shows up as the delta between them:

- `database/db-provisioning-preview-first` — first deploy (`netlify deploy`, no `--prod`) provisions the DB and applies migrations; promote only after verifying the draft.
- `database/db-first-deploy-failure` — first-deploy `401 Access Denied` on `createSiteDatabase` → retry preview-first, then surface and stop. Asserts the agent does **not** reach for `netlify api`, a raw `curl` to the API, on-disk tokens, or `psql`.
- `blobs/access-and-scoping` — surfaces the public-vs-private decision (or proceeds with the documented private default), uses a site-scoped `getStore`, and treats the serving layer as the access gate.
- `identity/dashboard-handoff` — hands dashboard-only setup (enable Identity, add the OAuth provider via "Use Netlify's app") to the user; asserts no API/token workaround to flip toggles.
- `frameworks/version-pinning` — does not guess package versions; checks `npm view <pkg> version` or installs without pins.

Branch brought up to date with `main`.


skills/CLAUDE.md
```diff
@@ -49,3 +49,26 @@ Read `netlify-deploy/SKILL.md` for the full deployment workflow — authenticati
 - Never hardcode secrets — use Netlify environment variables
 - Add `.netlify` to `.gitignore`
 - For framework-specific patterns, check the framework reference before writing custom Netlify Functions — the adapter may handle it
+
+## Use only documented Netlify CLI surfaces
```
jmarlena, line 53: Are we referencing how the agent should access "documented Netlify CLI surfaces"? Is this through the Netlify MCP Server the CLI reference site? Something else?

skills/netlify-identity/SKILL.md
```diff
@@ -9,16 +9,69 @@ Netlify Identity is a user management service for signups, logins, password reco
...
+- Autoconfirm (for prototypes) or email confirmation (for production)?
+- Public signup or invite-only?
+
+**If you don't have preferences here, tell me what you want overall and I'll pick sensible defaults** — typically email/password + Google OAuth, autoconfirm ON, registration Open for a prototype.
+
+Asking these *after* coding causes rework — both the auth UI shape and the dashboard config fall out of these answers.
+
+## When something fails, surface and stop
+
+If a deploy fails, an Identity callback 404s, an OAuth flow doesn't return, or `/.netlify/identity/*` is unreachable — report the failure to the user with the deploy log URL, the exact error, and the site URL, then stop. Do not curl the Netlify API to "fix" the Identity instance, do not invent recovery commands, do not bypass the dashboard. Identity instance state has no public API to repair — the recovery is always a dashboard handoff.
```
jmarlena: I love that you're thinking about agent failure recovery strategies here.

What exactly do we mean by "the recovery is always a dashboard handoff." Do we mean the fix is in the UI for the end user to do? Something else?

skills/netlify-identity/SKILL.md
```diff
@@ -9,16 +9,69 @@ Netlify Identity is a user management service for signups, logins, password reco
...
+- [ ] External providers → Add Google (etc.) with "Use Netlify's app"
+
+Tell me when these are flipped and I'll run the production deploy.
+```
+
+## Before you build
+
+If the prompt didn't already specify, ask the user a few short questions before scaffolding any auth code — the answers shape both the dashboard config above and the auth UI you'll write:
+
+- Email/password only, OAuth only, or both?
```
jmarlena: I wonder if we can give this question a little more context. Also is this for the entire app, a specific path or section of the site/app?

skills/netlify-identity/SKILL.md
```diff
@@ -9,16 +9,69 @@ Netlify Identity is a user management service for signups, logins, password reco
...
+- Which OAuth providers (Google, GitHub, GitLab, Bitbucket, Facebook)?
+- Autoconfirm (for prototypes) or email confirmation (for production)?
```
jmarlena: More context helpful here for end user. I wonder what email confirmation means exactly in this context. Also, should the next question go first?

skills/netlify-identity/SKILL.md
```diff
@@ -9,16 +9,69 @@ Netlify Identity is a user management service for signups, logins, password reco
 
 **Always use `@netlify/identity`.** Never use `netlify-identity-widget` or `gotrue-js` — they are deprecated. `@netlify/identity` provides a unified, headless TypeScript API that works in both browser and server contexts (Netlify Functions, Edge Functions, SSR frameworks).
 
+## Dashboard configuration (the agent does not own this)
```
jmarlena: Are we saying that the end user must turn on Identity first, and then the agent can help with configuration? 🤔

skills/netlify-frameworks/references/tanstack.md
```diff
@@ -2,6 +2,8 @@
 
 ## Setup
 
+> **Check current versions before pinning.** Knowledge cutoffs lag behind npm, and guessing a version tends to fail (`npm install` rejects it, or worse, installs something incompatible). Before pinning `@netlify/vite-plugin`, `@tanstack/react-start`, `vite`, or any other package in `package.json`, run `npm view <pkg> version` to get the current `latest`. Or omit explicit pins and let `npm install` pick them up.
```
jmarlena, line 5: I wonder if we can lead with the explicit instructions here first.

skills/netlify-database/SKILL.md
```diff
@@ -79,7 +104,32 @@ If you'd rather wire things up by hand, install the package directly:
...
+2. **User verifies on the draft URL** — and completes any dashboard-only setup along the way (e.g., enabling Identity if the project also uses it; see `netlify-identity/SKILL.md`).
+3. **Promote: `netlify deploy --prod`**.
+
+Why preview-first matters: the preview deploy provisions the database and applies migrations exactly the way the production deploy will, so a failure during preview is recoverable without prod ever entering a half-configured state. `--prod`-first works in the happy case but is harder to recover from when something goes wrong.
+
+### If the first deploy fails to provision the database
+
+**Symptom:** the build (or the Netlify Database setup extension inside the build) fails with a `401 Access Denied` on `createSiteDatabase`, typically on the very first deploy of a brand-new site. The deploy log shows the failure inside the extension's setup step.
```
jmarlena, line 121: Longer term I think we should invest in improving error messages here. This makes me wonder how much these skills are trying to compensate for lack of clarity and coverage in error messages.

skills/netlify-database/SKILL.md
```diff
@@ -19,6 +19,16 @@ Use **Netlify Database** for anything dynamic:
 
 Use **Netlify Blobs** only for **file and asset storage**: images, documents, exports, uploads, cached binary artifacts. Do not use Blobs as a dynamic data store — reach for Database instead. See `netlify-blobs/SKILL.md`.
 
+## Before you build
+
+If the prompt didn't already specify, ask the user a few short questions before scaffolding any database code — answers shape the schema, the seed data, and the query layer:
```
jmarlena, line 24: I wonder if we want to be more explicit about the format of questions to ask end users about their setup preferences.

codex/skills/netlify-identity/SKILL.md
```diff
@@ -9,16 +9,69 @@ Netlify Identity is a user management service for signups, logins, password reco
 
 **Always use `@netlify/identity`.** Never use `netlify-identity-widget` or `gotrue-js` — they are deprecated. `@netlify/identity` provides a unified, headless TypeScript API that works in both browser and server contexts (Netlify Functions, Edge Functions, SSR frameworks).
 
+## Dashboard configuration (the agent does not own this)
```
jmarlena: What do mean by the agent does not own this? Is this info for the agent to tell the user what to do? How might we put that into a user handoff step? Maybe we lead with steps signaling what the agent does and what the end user must do.

skills/netlify-identity/SKILL.md
```diff
@@ -9,16 +9,69 @@ Netlify Identity is a user management service for signups, logins, password reco
...
+### Recommended settings per use case
+
+| Use case | Registration | Autoconfirm | External providers |
+|---|---|---|---|
+| Prototype / demo | Open | ON | as requested |
+| Production with email signup | Open or Invite per product | OFF (real email confirmation) | configured with custom email templates / SMTP as needed |
+
+### Handoff checklist
+
+When the dashboard work is needed, give the user a copy-pasteable checklist **between the draft deploy and the production deploy** — not after the prod deploy fails:
```
jmarlena, line 44: This ordering could be helpful to capture in sequenced steps.

skills/netlify-identity/SKILL.md
```diff
@@ -9,16 +9,69 @@ Netlify Identity is a user management service for signups, logins, password reco
...
+
+```
```
jmarlena, line 46: I like that you're sharing a specific output to share with the end user.

jmarlena (commented):
Left some non-blocking comments for your review here! 💫 

I think the meta-challenge here is all the different setup pathways possible. I attempted to use the [skill-validator](https://github.com/agent-ecosystem/skill-validator/tree/main) here and it found this skill goes over the recommended token usage. I wonder if the following changes in structure might help reduce token usage here.

Formatting-wise, I wonder if we can save on token costs by adopting more meta-structure like these examples with numbered steps and a troubleshooting section and even leveraging a reference folder for specific scenarios and longer output formatting examples in an assets folder.

image

Example A: https://github.com/mongodb/agent-skills/blob/main/skills/mongodb-mcp-setup/SKILL.md
- Consistent wording for steps
- Shows steps, including sub-steps, which makes it easy to understand the relationships between steps and dependencies, which parts don't apply, what to skip
- Troubleshooting section at bottom so it doesn't get glossed over

Example B: https://github.com/mongodb/agent-skills/blob/main/skills/mongodb-query-optimizer/SKILL.md#example-workflow-1-help-with-specific-query
- Includes example workflows

Example C: https://github.com/elastic/elastic-docs-skills/blob/main/skills/review/docs-check-style/SKILL.md
- includes more metadata at top such as sources
- includes references at the bottom
- includes "You are [role]" with meta-task defined and the emphasis to not auto-fix. 

I've also read here that the beginning and end of a skill are most referenced so perhaps we can prioritize compatability info at top and troubleshooting at the bottom. 

As always, I think the best way forward here is iteration and more testing with different scenarios. I get this is extra hard with all the setup scenarios covered here! 🚀

PR #44, opened 2026-06-05
feat: add netlify-mcp-servers skill

seancdavis:
## What

Adds a new `netlify-mcp-servers` skill teaching developers to build, deploy, and secure MCP (Model Context Protocol) servers on Netlify — and integrate them into existing apps.

The existing [write-mcps-on-netlify guide](https://developers.netlify.com/guides/write-mcps-on-netlify/) leads with the SDK's Node-flavored `StreamableHTTPServerTransport` (plus a `fetch-to-node` bridge) in a way that breaks against some clients. This skill captures a simpler setup verified to work today.

## Approach (verified)

- **Official MCP SDK (`@modelcontextprotocol/sdk`, unpinned — install resolves to ≥ 1.25.0) + `zod`**, running the Web-standard `WebStandardStreamableHTTPServerTransport` in **stateless mode** (`sessionIdGenerator: undefined`, `enableJsonResponse: true`) inside a single Netlify Function.
- A Netlify Function already receives a Web `Request` and returns a Web `Response` — exactly what this transport speaks — so there's **no version pin, no `fetch-to-node` bridge, and no Node `req`/`res` adapter**. The native Web transport is what the SDK gained in 1.25.0; older guides reach for the Node-flavored `StreamableHTTPServerTransport` plus a `fetch-to-node` shim, neither of which is needed on Netlify. (This replaces the earlier pinned-to-`1.24.0` approach from before commit `b1a95ae0`.)
- The transport returns **HTTP 406** to any POST whose `Accept` header lacks *both* `application/json` and `text/event-stream` — but that's an MCP-spec requirement the *client* must satisfy, not a version-specific quirk (the same check exists in 1.24.0).
- Letting the SDK own JSON-RPC framing and the protocol handshake (no hand-rolled dispatch).

## Scope

- **Two auth models** — single shared secret (personal/single-user) vs per-user API keys backed by Netlify Identity (hash-not-plaintext, revocation, per-request user context).
- **Safety & permissions** — least privilege, reads vs writes, omitting destructive tools, confirmation-in-tool-descriptions + a real kill switch, separating client and backend credentials.
- **Native-first client connection** (Claude Code, Cursor, Claude Desktop Custom Connectors) with `mcp-remote` as the fallback.
- **Presigned file uploads** to Netlify Blobs.

## Structure

- `SKILL.md` (164 lines) — the spine, with the load-bearing transport setup inline.
- `references/authentication.md` — per-user API keys in depth.
- `references/connecting-clients.md` — full client matrix + OAuth/Custom Connectors deep-dive.
- `references/file-uploads.md` — the presigned-URL → Blobs flow.
- Router entry added to `skills/CLAUDE.md`.

## Generated files

The source of truth is `skills/` (plus the `axis-scenarios/` test scenarios). The `cursor/` and `codex/` files in the diff are **generated mirrors of `skills/`** — CI rebuilds and commits them on this PR so they travel in the same commit as the source, which is why they show up here. They're never hand-edited; edit `skills/` and let CI regenerate. #60 consolidates this into a single workflow and marks the mirrors as generated (`linguist-generated` + CodeRabbit exclusion) so they collapse in review and stop adding noise — once it lands, this diff gets quieter.


skills/netlify-mcp-servers/SKILL.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Stack
+
+Use the official MCP SDK with its Streamable HTTP transport, running statelessly inside a Netlify Function.
+
+```bash
+npm install @modelcontextprotocol/sdk@1.24.0 fetch-to-node zod
+```
+
+**Pin the SDK to `1.24.0`.** Versions `1.25+` tightened the Streamable HTTP transport so it rejects (HTTP 406) any POST whose `Accept` header doesn't contain *both* `application/json` and `text/event-stream`. Some clients don't send both, so the stricter versions break connections that `1.24.0` handles fine. This is worth re-checking over time as the SDK evolves — but pin for now so things just work. Letting the SDK own the protocol also means you don't hand-maintain JSON-RPC framing or the protocol-version handshake.
```
jah2488: 🤖 Did some source-diffing on this and I think the rationale is off, though the pin itself might still be the right call.

The both-Accept 406 is not new in 1.25. The same check (identical condition and identical error message) is already in the 1.24.0 pin target: `streamableHttp.js:323` rejects with 406 when the Accept header lacks either `application/json` or `text/event-stream`, and the identical check lives in 1.25.0 at `webStandardStreamableHttp.js:364`. It's unconditional in both (not gated by `enableJsonResponse`), so 1.24.0 also 406s a client that only sends one Accept value.

What actually changed in 1.25 is structural: the Node transport became a thin wrapper over a new `WebStandardStreamableHTTPServerTransport` via `@hono/node-server` (`streamableHttp.js:9,52`). That rewrite seems a more likely culprit against the `fetch-to-node` bridge. What did you see break on 1.25+? If it's the rewrite, worth re-wording so nobody unpins on a reason that doesn't hold (and the both-Accept requirement is really an MCP-spec thing the client has to satisfy, on any version).

-- Drafted and Posted by Claude Code with consent

skills/netlify-mcp-servers/SKILL.md
```diff
@@ -0,0 +1,162 @@
...
+import { timingSafeEqual } from "node:crypto";
+
+export function checkBearer(req: Request): boolean {
+  const expected = Netlify.env.get("MCP_BEARER_TOKEN");
+  if (!expected) return false;
+  const match = req.headers.get("authorization")?.match(/^Bearer\s+(.+)$/i);
+  if (!match) return false;
+  const a = Buffer.from(match[1]);
+  const b = Buffer.from(expected);
+  return a.length === b.length && timingSafeEqual(a, b); // avoid leaking length-timing
```
jah2488: 🤖 Tiny one: the `a.length === b.length &&` short-circuits before `timingSafeEqual`, so it does leak length-timing (harmless here since the token is fixed-length, but the comment claims the opposite). The actual reason the length check is needed is that Node's `timingSafeEqual` throws a `RangeError` ("Input buffers must have the same byte length") on unequal-length inputs. Might reword the comment to that.

-- Drafted and Posted by Claude Code with consent

skills/netlify-mcp-servers/SKILL.md
```diff
@@ -0,0 +1,162 @@
...
+
+Use the official MCP SDK with its Streamable HTTP transport, running statelessly inside a Netlify Function.
+
+```bash
+npm install @modelcontextprotocol/sdk@1.24.0 fetch-to-node zod
+```
+
+**Pin the SDK to `1.24.0`.** Versions `1.25+` tightened the Streamable HTTP transport so it rejects (HTTP 406) any POST whose `Accept` header doesn't contain *both* `application/json` and `text/event-stream`. Some clients don't send both, so the stricter versions break connections that `1.24.0` handles fine. This is worth re-checking over time as the SDK evolves — but pin for now so things just work. Letting the SDK own the protocol also means you don't hand-maintain JSON-RPC framing or the protocol-version handshake.
+
+## The server function
```
jah2488, line 35: 🤖 Possibly a follow-up rather than a blocker for this PR: since Netlify Functions already hand you a Web `Request` and take a Web `Response`, the `WebStandardStreamableHTTPServerTransport` that the 1.25+ SDK is built on speaks that natively. So on a current SDK you might be able to drop `fetch-to-node` and the Node req/res bridge entirely and run unpinned. Haven't tested it on Netlify, just flagging the direction.

-- Drafted and Posted by Claude Code with consent

seancdavis:
AXIS runs:

```
  AXIS — 7 scenarios · 1 agent
  ──────────────────────────────────────────────────
  mcp-servers/add-to-existing-app
    ✓ codex @no-context         77 / 100       (120.9s) 379,315 tok
    ✓ codex @with-skill         98 / 100       (251.4s) 648,942 tok

  mcp-servers/auth-single-secret
    ✓ codex @no-context         86 / 100       (43.0s) 59,013 tok
    ✓ codex @with-skill         98 / 100       (23.1s) 54,374 tok

  mcp-servers/build-server
    ✓ codex @no-context         62 / 100       (115.3s) 212,065 tok
    ✓ codex @with-skill         98 / 100       (185.5s) 420,687 tok

  mcp-servers/connect-client
    ✓ codex @no-context         90 / 100       (67.7s) 52,945 tok
    ✓ codex @with-skill         95 / 100       (66.0s) 115,947 tok

  mcp-servers/file-uploads
    ✓ codex @no-context         52 / 100       (43.0s) 40,599 tok
    ✓ codex @with-skill         92 / 100       (40.9s) 99,940 tok

  mcp-servers/per-user-api-keys
    ✓ codex @no-context         65 / 100       (35.1s) 75,652 tok
    ✓ codex @with-skill         98 / 100       (375.5s) 1,159,034 tok

  mcp-servers/safety-omit-destructive
    ✓ codex @no-context         85 / 100       (19.5s) 24,141 tok
    ✓ codex @with-skill         87 / 100       (39.3s) 102,574 tok

  ──────────────────────────────────────────────────
```

Gemini was giving me trouble, so I killed it. And about out of Claude tokens so I skipped it.

seancdavis:
Made some changes to address Justin's comments. Most recent run looks good!

```
  AXIS — 7 scenarios · 1 agent
  ──────────────────────────────────────────────────
  mcp-servers/add-to-existing-app
    ✓ codex @no-context         78 / 100       (136.4s) 358,755 tok
    ✓ codex @with-skill         98 / 100       (265.4s) 520,918 tok

  mcp-servers/auth-single-secret
    ✓ codex @no-context         82 / 100       (37.8s) 33,507 tok
    ✓ codex @with-skill         97 / 100       (31.5s) 58,532 tok

  mcp-servers/build-server
    ✓ codex @no-context         70 / 100       (140.7s) 320,823 tok
    ✓ codex @with-skill         97 / 100       (208.2s) 402,143 tok

  mcp-servers/connect-client
    ✓ codex @no-context         97 / 100       (75.7s) 55,978 tok
    ✓ codex @with-skill         96 / 100       (79.8s) 128,166 tok

  mcp-servers/file-uploads
    ✓ codex @no-context         49 / 100       (54.1s) 72,327 tok
    ✓ codex @with-skill         94 / 100       (38.4s) 95,128 tok

  mcp-servers/per-user-api-keys
    ✓ codex @no-context         61 / 100       (53.1s) 76,746 tok
    ✓ codex @with-skill         97 / 100       (391.7s) 1,400,445 tok

  mcp-servers/safety-omit-destructive
    ✓ codex @no-context         89 / 100       (24.3s) 24,047 tok
    ✓ codex @with-skill         94 / 100       (52.0s) 142,194 tok

  ──────────────────────────────────────────────────
  Average AXIS Result: 86 / 100
```

domitriusclark (commented):
Solid skill — scoping, auth, and safety guidance are good, and I verified the technical core: the `WebStandardStreamableHTTPServerTransport` import resolves correctly and ships in `@modelcontextprotocol/sdk` ≥ 1.25.0 (latest 1.29.0), so the "no version pin" advice holds for a fresh install. Cross-skill links resolve, the example code (`checkBearer`, HMAC `verifyUploadToken`, key-hashing flow) is correct, and the `add-to-existing-app` scenario matches the `nextjs-blog` fixture's `lib/posts.ts`.

Two non-blocking cleanups:

**1. The PR description is stale.** The "Approach (verified)" section still describes the pre-pivot setup — SDK pinned to `1.24.0` + `fetch-to-node` + Node `StreamableHTTPServerTransport`. Commit `b1a95ae0` dropped exactly that in favor of the native Web transport, which is what `SKILL.md` and all 7 scenarios now teach. Worth updating the body so "verified" points at the approach you actually shipped (nice detail to add: 1.25.0 introduced *both* the 406 Accept-header check and the web transport, which is why moving to ≥1.25.0 beats pinning to 1.24.0).

**2. Generated files are committed.** The description says "Only `skills/` is touched — `cursor/` and `codex/` are left to CI to regenerate," but the regenerated `cursor/rules/*` and `codex/*` files (incl. an unrelated `netlify-functions` codex refresh that fixes pre-existing drift on main) are in the diff. Harmless since CI is idempotent, but either drop them or reconcile the description.

---
🤖 Generated with Claude Opus 4.8 via Claude Code

seancdavis:
The cursor and codex rules are still in the pipeline. I'm looking to change this and will post in the channel, but they are still getting auto-generated at this time.

PR #58, opened 2026-06-25
feat(skills): clarify Netlify Identity vs access control, harden OAuth guidance

seancdavis:
Addresses the Claude → Netlify auth friction reported in **AX-61** (Identity vs Secure Access overlap, Claude scaffolding OAuth from scratch, and the perimeter + in-app identity double login).

These are the in-repo **skills** fixes. Out-of-repo tooling/product ideas (MCP read surface for auth config, SSO passthrough) are tracked as discussion on the Linear issue, not here.

## What changed

**New skill: `netlify-access-control`** — a decision-guide that disambiguates the three layers people conflate:
- **Netlify Identity** — app end users (`nf_jwt`), all plans.
- **Password Protection** — site-visitor perimeter (Basic = Pro+, team login = Enterprise).
- **Team/Org SAML SSO** — Netlify dashboard login (Enterprise).

It explains the Google "same button, different system" confusion, gives a decision tree that routes to **invite-only Identity** (just my team) / **Basic password** (soft gate) / **Auth0 extension** (whole company, single sign-on) / **two-layer** (genuine perimeter + Identity), and names the double login as a real, unbridgeable limitation. `references/two-layer-pattern.md` carries the per-option tradeoffs, the team-seat cost of team-login, and the visibility gap.

**Hardened `netlify-identity`:**
- Forbid scaffolding raw third-party OAuth when Identity is the goal; default to the built-in "Use Netlify's app" providers — this is the most frequent source of rework in the report.
- Added an Identity-vs-Secure-Access disambiguation callout near the top.
- Added a perimeter-vs-app pre-flight question (with the defaults outlet).
- Framed `getSettings()` as the authoring-time visibility workaround (don't hard-code providers).

**Router (`skills/CLAUDE.md`):** split app-level user auth from site access control and point both at the new skill.

## Notes
- Only `skills/` is touched. `cursor/` and `codex/` are regenerated by CI — not edited here.
- Authored via the skill-creator guidance; new skill kept tight with detail pushed to `references/`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified Netlify auth and access-control guidance across Identity, password protection, and SSO.
  * Added a new access-control reference explaining when to use each login layer and the tradeoffs of combining them.
  * Updated Identity guidance to better separate app logins from site-wide restrictions and to improve OAuth/setup recommendations.
  * Refined function configuration guidance for region and resource settings, including default behavior and plan limits.
  * Added scenario coverage for common access-control cases to improve consistency in future guidance.

PR #59, opened 2026-06-26
ci: stop committing generated cursor/codex output into PR branches

seancdavis:
## Problem

Both `build-cursor-rules.yml` and `build-codex-skills.yml` commit regenerated output back to the branch on **`pull_request`** events:

```yaml
if: github.event_name == 'push' || github.event_name == 'pull_request'
```

On any PR that touches `skills/**`, the bots push `cursor/rules/**` and `codex/**` commits onto the contributor's branch. That:

- **bloats every skills PR diff** with generated mirrors and pulls reviewers (e.g. CodeRabbit) onto generated files instead of the source;
- makes the **two jobs race** to push to the same branch (the "rerun the loser" flakiness);
- forces contributors to **pull before every push**.

This was surfaced in review on #58, where CodeRabbit commented on generated `codex/`/`cursor/` mirrors and flagged the churn.

## Change

Gate the **commit** step to `push` (main) only. The build still runs on PRs (so a broken `build-*.sh` is still caught), it just doesn't commit. Generated output is regenerated and committed when the change lands on `main` — which matches what the root `CLAUDE.md` already documents ("rebuilds these files whenever `skills/` changes on main").

## Effect

- PR diffs for skills changes are **source-only** (`skills/**`).
- No more cross-job push races on PR branches.
- `cursor/rules/**` and `codex/**` on `main` stay current via the push-to-main build.

Net: one-line condition change per workflow (plus an explanatory comment). No change to the build scripts or generated output format.

### DIFF
--- a/.github/workflows/build-codex-skills.yml
+++ b/.github/workflows/build-codex-skills.yml
@@ -27,8 +27,12 @@ jobs:
       - name: Build Codex skills from source
         run: bash scripts/build-codex-skills.sh
 
+      # Only commit regenerated output on push to main. On PRs the build still
+      # runs above (to catch a broken script), but committing here is skipped so
+      # generated files stay out of PR diffs and the cursor/codex jobs don't race
+      # to push to the contributor's branch.
       - name: Commit updated Codex skills
-        if: github.event_name == 'push' || github.event_name == 'pull_request'
+        if: github.event_name == 'push'
         run: |
           git config user.name "github-actions[bot]"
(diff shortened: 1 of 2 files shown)


seancdavis:
Superseded by the consolidated, parity-preserving approach. Committing generated output only on main (as this PR did) would let a release-please version bump tag a commit before cursor/codex regenerated — the parity race surfaced in review. Replaced by a single workflow that commits the mirrors alongside source on the PR, so they always travel in the same commit. See the new PR.

PR #60, opened 2026-06-26
ci: consolidate cursor/codex builds into one parity-preserving workflow

seancdavis:
Supersedes #59. That PR proposed committing generated output *only on main*, which would let a `release-please` version bump tag a commit before `cursor/`/`codex/` regenerated (the parity/ordering race raised in review). This takes the structural approach instead: the mirrors always travel in the same commit as the source.

## How it behaves, end to end

1. **Generation still runs on PRs.** When a PR changes `skills/**` (or the build scripts / this workflow), CI rebuilds `cursor/` and `codex/` and commits them back onto the PR branch — the mirrors travel with the source.
2. **The mirrors are collapsed in review, not removed.** `.gitattributes` marks `cursor/**` and `codex/**` `linguist-generated`, so GitHub folds them by default in the Files-changed tab (expandable, still there) and drops them from language stats.
3. **CodeRabbit skips them.** `.coderabbit.yaml` `path_filters` exclude `cursor/**` and `codex/**`, so review feedback lands on `skills/`.
4. **Versioning stays in parity.** `main` is squash-only, so each PR lands as a single commit titled with its (conventional, lint-enforced) PR title — and the bot's "rebuild generated" commits get squashed away. `release-please` (on push to `main`) opens/updates a release PR and cuts the version when that PR merges. Because the mirrors ride in the *same* squash commit as the source, every `main` commit is self-consistent, so the commit release-please tags always has `cursor/`/`codex/` in parity with `skills/` — closing the stale-mirror race from #59.

Net: edit `skills/` only, reviews stay on `skills/`, and `main` is always in parity without a version-bump race.

## What this does

Replaces the two separate build workflows (`build-cursor-rules.yml`, `build-codex-skills.yml`) with a single **`build-generated-outputs.yml`** that builds `cursor/` and `codex/` and commits them in **one commit**.

**Parity is structural, not timing-based.** Because the mirrors are committed alongside the source change on the PR branch, they land in the same merge commit as `skills/`. Every commit on `main` is self-consistent, so whatever commit the `release-please` PR tags is in parity — no window where a version bump ships stale generated output.

**No inter-job race.** One job → one push replaces two workflows that each pushed to the branch and raced (the "rerun the loser" flakiness). A `concurrency` group serializes runs per branch as a backstop.

## Behavior by event

| Event | Checkout | Action |
|---|---|---|
| push to `main` | branch (writable) | rebuild + commit mirrors |
| same-repo PR | branch (writable) | rebuild + commit mirrors back to the PR |
| fork PR | PR merge ref (read-only) | rebuild + **fail** if mirrors are stale (forks can't be pushed to; they're regenerated on merge to `main`) |

The bot commit carries **no skip-CI marker** — the workflow's `paths:` filter already excludes `cursor/**` and `codex/**`, so a generated-only commit can't retrigger the build, and dropping the marker lets normal PR checks evaluate the actual head commit.

## Review / diff noise

Instead of removing the mirrors from PRs (which would break parity), they're de-emphasized:
- **`.gitattributes`** marks `cursor/**` and `codex/**` `linguist-generated`, so GitHub collapses them in diffs.
- **`.coderabbit.yaml`** excludes them from review, so feedback lands on the source.

## Files

- ➕ `.github/workflows/build-generated-outputs.yml`
- ➖ `.github/workflows/build-cursor-rules.yml`, `.github/workflows/build-codex-skills.yml`
- ➕ `.gitattributes`, `.coderabbit.yaml`
- ✏️ `README.md`, `CONTRIBUTING.md`, `CLAUDE.md` (document the PR-and-main model; note fork PRs verify rather than auto-commit)

## Notes

- **Merge order:** land this before #58 so #58's merge to `main` runs the single-commit workflow rather than the old racing pair. (#44 already merged under the old racing pair — harmless, since the builds are idempotent.)
- The fork-PR path is logically correct and the workflow runs green, but it isn't exercised end-to-end here (no fork PR involved).
- If "Build generated outputs" later becomes a *required* status check, note that a mirror-only bot commit won't produce that check (path filter) — use a path-agnostic gate rather than requiring it on those commits.


seancdavis:
Addressed all three in `8be2914`.

**P1 — `[skip ci]` removed.** Agreed — it only had the downside (skipping the normal checks on the real head) with no upside: the workflow's `paths:` filter already excludes `cursor/**` and `codex/**`, so a generated-only commit can't retrigger the build regardless. PR checks now evaluate the actual head commit.

**P2 — fork / read-only-token PRs handled.** The commit-back step is now gated on a same-repo head (`github.event.pull_request.head.repo.full_name == github.repository`). Fork PRs instead run a verify step that rebuilds and **fails with a clear diff message** if the mirrors are stale, rather than erroring on `git push` — and they're regenerated/committed when the change merges to `main`. (Note: Dependabot/Renovate PRs here bump deps in `package.json`, which don't match the `skills/**` path filter, so they don't trigger this workflow at all.)

**P3 — docs fixed.** `README.md`, `CONTRIBUTING.md`, and `CLAUDE.md` now say the mirrors rebuild on PRs and on push to `main`, committed alongside the change.

Heads-up for when you wire up branch protection: with `[skip ci]` gone, a generated-only bot commit still won't produce a "Build generated outputs" run (path filter), so if you make that a *required* check you'd want a path-agnostic gate rather than requiring it on mirror-only commits. Not changed here — just flagging.

seancdavis:
Fixed all three in `e623327` — CI is green on this head now (run 28251424020).

**P1 — skip token in the commit message.** 😅 The previous commit's message literally contained the skip marker (it was *describing* removing it), so Actions skipped that head. No force-push — this commit re-touches the workflow with a clean message, so the real head ran and passed.

**P2 — fork checkout.** Split into two conditional checkouts: push + same-repo PRs use `ref: github.head_ref || github.ref` (writable branch, so the mirrors can be committed back); fork PRs use the default `pull_request` merge-ref checkout (this repo's scripts + the fork's changes, read-only), so the fork verify step actually runs instead of failing at checkout.

**P3 — docs.** README, CONTRIBUTING, and CLAUDE now say same-repo PRs auto-commit while fork PRs must include regenerated output (or leave it for a maintainer).

Note: the fork path isn't exercised by this PR (no fork involved), but the same-repo/push path ran green on this head.

PR #64, opened 2026-06-29
fix(axis): reconcile Drizzle migration timestamp prefix with skill

seancdavis:
## Summary

Reconciles a contradiction between the Drizzle config scenario and the database skill. The judge required `migrations: { prefix: 'timestamp' }`, but the skill's canonical `drizzle.config.ts` omits the `migrations` block entirely (timestamp is already the `drizzle-kit@beta` default). An agent copying the skill verbatim therefore **failed** the scenario. (AX-76, part of AX-29.)

## Changes

1. **`axis-scenarios/database/db-drizzle-config.ts`** — the prefix check is now tolerant: PASS when the config omits the `migrations` block (relying on the timestamp default, as the canonical skill example does) **or** sets `prefix: 'timestamp'` explicitly; FAIL only when it forces a sequential/index prefix (e.g. `prefix: 'index'`).
2. **`skills/netlify-database/SKILL.md`** — added a short note explaining why no `migrations` block is needed.

## Validation

Local AXIS run (codex + claude-code; gemini skipped), `database/db-drizzle-config`:

| variant | codex | claude-code |
|---|---|---|
| with-skill | 97 | 98 |
| no-context | 83 | 84 |

- The reconciled prefix check scored **10/10 in all four runs**, including the no-context runs that omit the `migrations` block — confirming the false-fail is gone.
- Strong, meaningful uplift (goalAchievement **67 → 100**): without the skill, both agents send migrations to the drizzle-kit default `drizzle/` (the Netlify deploy only applies from `netlify/database/migrations`) and add an unneeded `dbCredentials` connection block. The skill fixes both. The prefix check is correctly **neutral** across variants — it was a false-fail, not the discriminator.

## Notes for reviewers

- The tolerant judge is robust to the `@beta` drizzle-kit config surface shifting — it only fails a *forced* sequential prefix, so re-confirming the exact beta config keys isn't a merge blocker.
- Editing `skills/` triggered the `cursor/`/`codex/` rebuild (regenerated-output commit included).

Closes AX-76.

## Summary by CodeRabbit

* **Documentation**
  * Clarified guidance for database migrations so a migration block is no longer required.
  * Noted that timestamp-based migration names are acceptable and help avoid filename collisions.
  * Added a warning against sequential-style migration naming.


seancdavis:
AXIS run:

```
────────────────────────────────────────────────────────────────────────────────────────────────────
  Timestamp:  2026-06-30T14:00:45.807Z
  Duration:   69.0s
  Scenarios:  4 completed, 4 skipped (4 total)
  AXIS Result: 91 / 100 (average)

  Scenario               Agent                     AXIS    Goal    Env.    Svc.    Agent   Duration   Cost
  ────────────────────────────────────────────────────────────────────────────────────────────────────
  database/db-drizzle-config claude-code @with-skill   98      100     93      100     95      19.3s      $0.2556
  database/db-drizzle-config claude-code @no-context   84      67      93      100     95      21.7s      $0.1487
  database/db-drizzle-config codex @with-skill         97      100     93      100     93      25.9s      —
  database/db-drizzle-config codex @no-context         83      67      100     96      83      39.8s      —
  ────────────────────────────────────────────────────────────────────────────────────────────────────

```

PR #67, opened 2026-06-29
test(axis): activate skipped database scenarios

seancdavis:
## Summary

Resolves the two database scenarios that carried a bare `skip: true` (no rationale). Un-skipping them surfaced **why** they were skipped: both were unsatisfiable in the AXIS sandbox as written. So this PR fixes them rather than silently re-disabling. (AX-79, part of AX-29.)

## Changes

- **`db-inspect-tables`** — reframed the prompt to ask for *the commands to run*, not live output. The original *"I want actual output, not a script"* demanded a live query, but the sandbox has no authenticated Netlify Database — so agents (even with the skill) reached for `psql` + a DB URL to comply, the exact unsafe path the judge fails. It now tests the correct approach (`netlify database connect --query` against `information_schema`); the no-`psql` / no-DDL safety checks stay meaningful.
- **`db-production-data-change`** — added the `drizzle-db` fixture so there's a real Netlify-DB/Drizzle project to create the migration in. Without it, agents had nowhere to write the file and scored 0 on the "creates a migration file" checks. Also softened the over-specific preview→prod explanation check.
- Removed the bare `skip: true` from both.

## Validation

A re-run confirmed both scenarios now execute and pass cleanly — the `psql`-temptation and missing-file failures are resolved.

## Notes for reviewers

- This carries a copy of the `drizzle-db` fixture also added in AX-81 (#69) — **merge #69 first** for a clean identical-add.
- Touches only `axis-scenarios/` + `axis-fixtures/` — no `cursor/`/`codex/` rebuild.

Closes AX-79.


seancdavis:
```
  AXIS — 2 scenarios · 2 agents
  ──────────────────────────────────────────────────
  database/db-inspect-tables
    ✓ codex @no-context         94 / 100       (53.9s) ~5,006 tok
    ✓ codex @with-skill         97 / 100       (20.3s) ~14,872 tok
    ✓ claude-code @no-context   79 / 100       (29.7s) ~8,222 tok
    ✓ claude-code @with-skill   98 / 100       (21.2s) ~8,726 tok

  database/db-production-data-change
    ✓ codex @no-context         98 / 100       (55.3s) ~23,969 tok
    ✓ codex @with-skill         98 / 100       (81.7s) ~43,550 tok
    ✓ claude-code @no-context   96 / 100       (181.4s) 315,091 tok
    ✓ claude-code @with-skill   97 / 100       (90.8s) ~33,703 tok

  ──────────────────────────────────────────────────
```

seancdavis:
Codex review: 2026-06-30 16:00:15 EDT

- Medium: `db-production-data-change` says the `categories` table already exists, but the shared `drizzle-db` fixture only defines the `items` table. That mismatch makes the scenario less realistic and may push agents toward adding schema/DDL even though the judge is trying to test a data-only production seed migration. Add a minimal `categories` table to the fixture, or give this scenario a dedicated fixture that matches its prompt.

PR #69, opened 2026-06-29
test(axis): add fixtures for existing-project scenarios

seancdavis:
## Summary

Adds three test fixtures so scenarios that assume an existing project stage a **real workspace** instead of running against an empty or mismatched one. Previously the only fixture was `nextjs-blog`. (AX-81, part of AX-29.)

## Changes

- New fixtures under `axis-fixtures/`:
  - **`astro-static`** — a minimal Astro project, deliberately **without** `@astrojs/netlify` so the scenario's "installs the adapter" check stays meaningful.
  - **`drizzle-db`** — a Netlify-DB + Drizzle project with `db/schema.ts` and a `drizzle.config.ts` pointing at `netlify/database/migrations/`.
  - **`edge-auth-site`** — a static site staging the `netlify/edge-functions/auth.ts` the scenario references, plus a minimal `netlify.toml`.
- Wired via `copyFixture` into **`frameworks/astro-adapter`**, **`database/db-schema-change`**, and **`config/headers-and-edge`** (the last previously told the agent to wire an edge function that didn't exist).
- Left `db-drizzle-setup`, `identity-event-signup`, and `agent-runner/check-task-status` fixture-free — they're genuinely greenfield or pure Q&A.

## Validation

Local AXIS run of the three wired scenarios (codex + claude-code; gemini skipped) — all execute cleanly against the fixtures, no failures: `astro-adapter` 88–96, `db-schema-change` 81–95, `headers-and-edge` 90–99.

## Notes for reviewers

- The `drizzle-db` fixture is also copied into AX-79 (#67) to fix `db-production-data-change`; identical files, so either merge order is clean.
- Touches only `axis-scenarios/` + `axis-fixtures/` — no `cursor/`/`codex/` rebuild.

Closes AX-81.


seancdavis:
Codex review: 2026-06-30 16:00:15 EDT

- Medium: The new `astro-static` fixture pins `astro` to `^4.15.0`, while the related Astro scenario/skill changes are moving to Astro 5 semantics (`hybrid` removed; `static`/`server` plus per-route prerender control). That weakens the fixture coverage because the staged project does not match the behavior under test. Use Astro 5 in the fixture, or omit the explicit pin if AXIS does not install dependencies during the scenario.

PR #72, opened 2026-06-29
test(axis): add runtime-primitive scenarios for functions, edge, caching, blobs, image

seancdavis:
## Summary

Adds 25 AXIS coverage scenarios for runtime primitives — Functions, Edge Functions, Caching, Blobs, and Image CDN — targeting the behaviors agents most often get subtly wrong. (AX-59, part of AX-29.)

## Scenarios added

**Functions (7):** `waitUntil` background work; `config.path` routing; default-route fallback; shared non-function `_shared/` helper modules; the `formSubmitted` platform event handler; scheduled functions running on published deploys only; request/response payload limits.

**Edge Functions (6):** `onError` bypass; method scoping; manual `Cache` API use; import maps; env access; edge-vs-serverless selection.

**Caching (6):** durable cache; stale-while-revalidate; `Vary` by cookie and query; deploy-triggered tag invalidation; `Cache-Status` debugging; basic-auth vs cache conflict.

**Blobs (5):** deploy-scoped store; strong consistency; JSON helpers; metadata-only reads; local-dev setup.

**Image CDN (1):** transform `position` / `quality` / `blurhash` params.

## Validation

Full local run (codex + claude-code, 100 runs): **100/100 completed, 0 failed, avg AXIS 94.** `caching/tags-deploy-invalidation` runs clean against the corrected `Netlify-Cache-Tag` skill (AX-74, merged).

Two scenarios were repaired during validation:

- **`functions-shared-helpers`** — reframed from "I have two functions…" to "Create two functions…". The old prompt assumed an existing `orders.ts`/`customers.ts` that wasn't staged, so codex found an empty workspace and built nothing (goal 17). Greenfield now has the agent create both functions plus the `_shared/` module — codex/claude both 98.
- **`blobs/json-helpers`** — the judge required importing `getStore`, but both agents reached for `getDeployStore`, which is equally valid (and arguably the better fit for a per-build snapshot). Relaxed to accept either — now 96–98.

`functions-form-event` (strict variant) held: codex produces the `formSubmitted` object handler reliably; claude-code missed the pattern on one run and cleared it on the next, so the scenario is correctly catching real agent variance rather than a bug.

## Notes

Judges assert the *documented* runtime behavior — `_shared/` helpers that aren't deployed as endpoints, deploy-scoped Blobs, published-only schedules, the event-handler object shape for form submissions — not hand-rolled approximations.

Closes AX-59.


seancdavis:
Codex QA review: No actionable findings. The AX-59 runtime-primitive scenarios are intentionally simple rubric additions, and the security-sensitive checks consistently reject token-file reads, direct API workarounds, unsafe secret access, and fake platform APIs. The PR body documents a broad local AXIS run plus targeted repairs for the previously weak scenarios.

PR #81, opened 2026-07-07
fix(skills): correct factual errors in caching, ai-gateway, deploy, forms, cli, functions, and frameworks skills

seancdavis:
## Summary

Fixes verified FACTUAL errors across seven skills (AX-103, report §A plus the two doc-currency items). Each wrong skill is corrected, and where a scenario encoded the same error, the scenario is fixed too so the rubric stays in parity with the skill.

## What changed and why

- **Caching — deploy invalidation (skills/netlify-caching/SKILL.md):** the skill claimed `Netlify-Cache-Tag` responses are excluded from automatic deploy-based invalidation. That is wrong: `Netlify-Cache-Tag` is purge-only and its responses are still wiped on deploy. `Netlify-Cache-ID` is the header that opts a response out of automatic deploy invalidation and auto-registers as a purge tag. Rewrote the section accordingly and added a worked `Netlify-Cache-ID` + `purgeCache({ tags: [sameId] })` example. Correspondingly rewrote `axis-scenarios/caching/tags-deploy-invalidation.ts`, whose checks were INVERTED (required `Netlify-Cache-Tag`, forbade `Netlify-Cache-ID`).
- **Caching — Cache-Status (skills/netlify-caching/SKILL.md):** replaced the bare `HIT`/`MISS`/`REVALIDATED` description with the real RFC 9211 format (named layers, `fwd=miss`, `ttl=…`).
- **AI Gateway (skills/netlify-ai-gateway/SKILL.md):** dropped the stale `-preview` suffix from the Gemini image IDs (`gemini-3-pro-image`, `gemini-3.1-flash-image`), added `gemini-3.1-flash-lite-image`, and fixed the two example usages. In the chat list, dropped `-preview` from `gemini-3.1-flash-lite` (docs no longer carry the preview form) while leaving `gemini-3-flash-preview`/`gemini-3.1-pro-preview`/`gemini-3.1-pro-preview-customtools`, which docs still list with `-preview`. Verified against docs.netlify.com/build/ai-gateway/overview/. Also updated the stale example IDs in `axis-scenarios/ai-gateway/image-generation.ts` and `image-to-image.ts`.
- **netlify.toml reference (skills/netlify-deploy/references/netlify-toml.md):** removed the fabricated `[[functions]]` array-of-tables (`path`/`function`) block — only `[functions]` and `[functions."name-or-glob"]` exist — and replaced it with a real per-function override example. Fixed `axis-scenarios/config/functions-directory.ts`, whose judge phrasing treated the fictional array-of-tables as real. Also corrected the `publish` comment: it resolves under `base` when `base` is set, not relative to repo root.
- **Forms (skills/netlify-forms/SKILL.md):** added Astro to the SSR list and a caveat that form detection parses prerendered HTML only, so a `data-netlify` form on an Astro on-demand/SSR route is never registered.
- **CLI (skills/netlify-cli-and-deploy/SKILL.md):** corrected the implication that pushing to a branch auto-creates a branch deploy — branch deploys are off by default and must be enabled.
- **Functions (skills/netlify-functions/SKILL.md):** fixed the dead `#resource-configuration` anchor to `#memory-or-vcpu`.
- **Frameworks (skills/netlify-frameworks/SKILL.md):** reconciled the TanStack Start detection signature to current (`vite.config.*` + `@tanstack/react-start`) instead of the outdated `app.config.*`.

Out of scope (owned by other issues): `process.env` guidance (AX-104) and cross-cutting-rule propagation (AX-105).

Closes AX-103

## Summary by CodeRabbit

* **Documentation**
  * Clarified AI Gateway image model examples and removed outdated preview-name references.
  * Updated caching docs to explain cache IDs, deploy persistence, purge behavior, and improved cache-status debugging guidance.
  * Refined deploy, functions, forms, frameworks, and configuration guidance for more accurate setup and behavior.

* **Bug Fixes**
  * Corrected scenario checks and examples to match current image model names and configuration patterns.
  * Improved guidance around branch deploys and form detection to reduce misconfiguration.

### DIFF
--- a/axis-scenarios/ai-gateway/image-generation.ts
+++ b/axis-scenarios/ai-gateway/image-generation.ts
@@ -8,7 +8,7 @@ export default {
   judge: [
     { check: "Uses the `@google/genai` package — image generation on the gateway is Gemini-only" },
     { check: "Does NOT use `@google/generative-ai` (older deprecated package) and does NOT attempt OpenAI image models like dall-e or gpt-image" },
-    { check: "Uses a Gemini image model (e.g. 'gemini-2.5-flash-image' or 'gemini-3-pro-image-preview')" },
+    { check: "Uses a Gemini image model (e.g. 'gemini-2.5-flash-image' or 'gemini-3-pro-image')" },
     { check: "Does NOT pass a custom apiKey or baseURL to the GoogleGenAI client — the gateway auto-configures it" },
     { check: "Does NOT read or set `GEMINI_API_KEY` / `GOOGLE_API_KEY` — those disable the gateway" },
     { check: "Extracts the image bytes from `response.candidates[0].content.parts[*].inlineData` (base64 string) rather than expecting a URL" },
--- a/axis-scenarios/ai-gateway/image-to-image.ts
+++ b/axis-scenarios/ai-gateway/image-to-image.ts
@@ -16,7 +16,7 @@ export default {
     },
     {
       check:
-        "Pins a Gemini IMAGE model (e.g. `gemini-2.5-flash-image`, `gemini-3.1-flash-image-preview`, or `gemini-3-pro-image-preview`).",
+        "Pins a Gemini IMAGE model (e.g. `gemini-2.5-flash-image`, `gemini-3.1-flash-image`, or `gemini-3-pro-image`).",
     },
     {
       check:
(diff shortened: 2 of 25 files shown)


seancdavis:
Codex review: I audited this against AX-103 and the branch diff. I did not find blocking issues. Generated Codex/Cursor mirrors are in sync, and `skill-validator check --strict skills/` passes with network access. I did not run AXIS tests per request.

PR #82, opened 2026-07-07
fix(axis): reconcile over-tested judge checks with skill facts (AX-104)

seancdavis:
## Summary

Reconciles the confirmed OVER-TESTING findings from the AXIS audit (report §B): judge checks that asserted platform facts no loaded skill documents. Per `axis-scenarios/README.md`, the preferred fix is to teach the fact in the skill; a few checks are relaxed only where there is no documentable fact.

## Skills — facts taught

- **netlify-functions**: `process.env` softened from "don't use it" to "prefer `Netlify.env.get` for cross-runtime/edge portability; `process.env` is also valid in Functions" (the Edge-only `Netlify.env.get` requirement is kept). The Region section now states region values are IATA airport codes and lists `cmh`/`dub`/`fra`. The Memory or vCPU section now documents the accepted `memory` value syntax (a number of MB like `2048`, or a string like `'2gb'`/`'2048mb'`, case-insensitive, 1024–4096 MB).
- **skills/CLAUDE.md**: same `process.env` softening on the general env-var rule; the Edge-only accessor rule is kept.
- **netlify-config**: the `[[edge_functions]]` example now shows `excludedPath` as a toml key (string or array of globs) carving exceptions out of `path`.
- **netlify-ai-gateway**: adds `GOOGLE_API_KEY` to the "don't set your own key" rule and notes the `@google/genai` SDK reads either Gemini var, so a user-set `GOOGLE_API_KEY` shadows the gateway.
- **netlify-caching**: Cache Key Variation now documents the enumerated `country=us|de` form (lowercase ISO codes) and the composite comma-separated form (e.g. `query=theme, cookie=plan`), plus why enumerating beats keying on a whole dimension like a bare `Vary: Cookie`.
- **netlify-frameworks/references/vite.md**: installs `@netlify/vite-plugin` with `-D` so the skill matches the scenario's dev-dependency check.

## Scenarios — checks relaxed / reframed

- **frameworks/astro-adapter, nextjs-api-route, nextjs-isr**: the `process.env`-penalizing checks are reframed to "don't hardcode secrets" — `process.env` is valid and idiomatic in Next.js/Astro route handlers.
- **caching/vary-by-country**: reframed to the enumerated `country=us|de|jp` form (bare `Netlify-Vary: country` is not documented); the prompt now targets a fixed set of markets so the intent matches what enumeration can express.
- **blobs/json-helpers**: drops the undocumented `getJSON` helper alternative, leaving the documented `store.get(key, { type: 'json' })`.
- **deploy/manual-deploy**: drops the undocumented `netlify sites:create` path; keeps the documented `netlify init --manual`.
- **identity/role-based-access**: drops the fabricated `user.roles` alternative; keeps the documented `user.app_metadata.roles`.

Closes AX-104

### DIFF
--- a/axis-scenarios/blobs/json-helpers.ts
+++ b/axis-scenarios/blobs/json-helpers.ts
@@ -7,7 +7,7 @@ export default {
     "At the end of each build we produce a single site-configuration snapshot — one JSON object with feature flags and theme settings — and we want to stash it in Netlify Blobs and read it back as a typed object at runtime. Use Netlify Blobs' JSON convenience helpers for the write and read.",
   judge: [
     { check: "Writes the snapshot with `store.setJSON(key, obj)` rather than manually `JSON.stringify`-ing into `store.set`" },
-    { check: "Reads it back as parsed JSON via `store.get(key, { type: 'json' })` (or a `getJSON` helper) — NOT a manual `JSON.parse(await store.get(key))`" },
+    { check: "Reads it back as parsed JSON via `store.get(key, { type: 'json' })` — NOT a manual `JSON.parse(await store.get(key))`" },
     { check: "Imports a store constructor from '@netlify/blobs' — either `getStore` or `getDeployStore` (deploy-scoped is a fine choice for a per-build snapshot)" },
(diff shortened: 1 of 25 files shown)


seancdavis:
Codex review: I audited this against AX-104 and the branch diff. I did not find blocking issues. GitHub already shows build and validate checks passing. I did not run AXIS tests per request.

PR #92, opened 2026-07-13
feat(ctx-pipeline): add Stage-2 agent-context receiver (AX-97)

seancdavis:
Stage 2 of the Context Pipeline (AX-97), the **receiving** side.

`netlify/docs` generates and AXIS-tests a skill per grouping upstream. This repo's job is deterministic distribution: pull that artifact and import it into `skills/` byte-for-byte. No model call, no content rewrite here — if we ever transform content, that's when this repo would add its own AXIS.

### What's in it

- **`scripts/ctx-receive.mjs`** — reads `.ctx-gen/config.json`, compares each grouping's `manifest.generation.source_hash` to the `.ctx-gen/state.json` cache, and on change copies `agent-context/<g>/skill/` → `skills/<name>/`. Guards against mapping drift (the generated `SKILL.md` must declare the name we map to).
- **`.ctx-gen/`** — config (grouping → skill mapping + `importerVersion`), the delta cache, and a README describing the receiving side.
- **`.github/workflows/ctx-pipeline-receive.yml`** — resolves the docs ref, imports, and opens a draft PR touching `skills/`, which fires the existing `validate-skills` + `build-generated-outputs` (cursor/codex parity) gates. Human reviews and merges; rollback = revert.

### Verified locally (against docs `d4389ce`)

- First import → **byte-for-byte identical** to the upstream `netlify-functions` skill
- Cache write, then a second run is a clean cache-hit no-op
- Stale cache → re-import + state rewrite

### Not wired yet (inert until then)

`netlify/docs` is private, and a `GITHUB_TOKEN`-authored PR won't trigger our checks, so the workflow needs two secrets — `DOCS_READ_TOKEN` (read docs) and `CTX_PIPELINE_PR_TOKEN` (push/PR so checks run) — plus the `CTX_PIPELINE` repo variable to arm automatic triggers. Manual `workflow_dispatch` works once the two secrets exist. Details in the workflow header and `.ctx-gen/README.md`.

Draft until we've run it end-to-end.

Ref: AX-97.

## Summary by CodeRabbit

* **New Features**
  * Added an automated pipeline to import updated documentation-based skills and propose changes through a draft pull request.
  * Added configuration for sourcing and grouping imported skills.
  * Added manual and event-driven pipeline triggers with serialized updates.
  * Added dry-run support and change tracking for imports.

* **Documentation**
  * Added documentation describing the context pipeline, its inputs, triggers, and update process.

* **Chores**
  * Updated ignored paths for generated documentation and pipeline-related files.


domitriusclark:
The safety-net cron pulls docs `main` directly, which can skip the quality gate: if context merges to docs main but the AXIS gate *fails*, we deliberately don't dispatch — but the Tuesday poll would import it anyway.

Suggestion: after a passing gate, the docs merge workflow can move a `ctx-gated` tag to the passing commit (small change on our side — happy to add it). Then the poll checks out that tag instead of `main`, and the safety net keeps the same guarantee as the dispatch path.

domitriusclark:
If a second docs change lands before the first import PR merges, run two compares against main's `state.json` (stale) and opens an overlapping PR — both touching the same skill dirs and state file. Rare, but when it happens it's merge-conflict cleanup. Would a single rolling branch (force-push + one standing PR) be simpler than one branch per docs SHA?

seancdavis:
Thanks @domitriusclark — both are fixed on the branch.

**Overlapping PRs:** went with your rolling-branch suggestion. One standing branch (`ctx-pipeline/agent-context-sync`) force-updated in place, and the job concurrency group serializes runs so there's a single writer — a second docs change updates the same PR instead of opening a colliding one. (09ea8b8)

**Cron skipping the gate:** good catch — and rather than add the `ctx-gated` tag, I just dropped the scheduled poll entirely. It was the only thing that needed the tag, so removing it kills that whole mechanism *and* the cross-repo dependency — no tag-move on your side after all. Automated imports are now dispatch-only (gated by construction, since docs only dispatches on a passing gate); a missed dispatch self-heals on the next one (each run imports the full current skill, not a delta) or a manual run recovers it. (f45baf7)

Appreciate the review.

domitriusclark (approved):

PR #95, opened 2026-07-17
feat(netlify-split-testing): add split testing skill

agroofficial1256-spec:
## What
Adds a new skill covering Netlify's Split Testing feature (branch-based A/B testing and private beta releases).

## Why
Split Testing wasn't covered by any existing skill in this repo.

## Changes
- New `skills/netlify-split-testing/SKILL.md` — requirements/limitations, how to run a test, the `nf_ab` cookie, and client-side analytics tracking (GA/Segment/snippet injection).
- Updated `skills/CLAUDE.md` with a routing entry pointing to the new skill.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for setting up Netlify Split Testing or private betas by routing traffic between deployed branches.
  * Documented setup and stopping procedures, visitor assignment, opt-in testing, analytics tracking, requirements, and limitations.

### DIFF
--- a/skills/CLAUDE.md
+++ b/skills/CLAUDE.md
@@ -44,6 +44,7 @@ Read `netlify-access-control/SKILL.md` to pick the right layer — app-level Ide
 Read `netlify-deploy/SKILL.md` — the single deploy skill covering CLI install/auth, site linking, Git-based continuous deployment (the primary path), manual/local and CI deploys, and deploy troubleshooting. (Environment variables live in `netlify-config`; local dev lives in `netlify-frameworks`.)
 
 **Setting up A/B testing or a private beta by splitting traffic between deployed branches?**
+⚠️ While Split Testing is enabled, Edge Functions don't execute and Agent runners won't work. Don't recommend it for sites relying on edge/SSR/agent features
 Read `netlify-split-testing/SKILL.md` for the branch-based Split Testing feature, the `nf_ab` cookie, client-side analytics tracking, and its conflicts with Edge Functions, Agent Runners, and Functions/proxy-generated content.
 
 **Want to run AI agents (Claude, Codex, Gemini) remotely on your site?**
--- a/skills/CLAUDE.md
+++ b/skills/CLAUDE.md
@@ -43,6 +43,9 @@ Read `netlify-access-control/SKILL.md` to pick the right layer — app-level Ide
 **Deploying a site to Netlify — installing/authenticating the CLI, linking a project, Git-based or manual deploys, CI deploys, or troubleshooting a failed deploy?**
 Read `netlify-deploy/SKILL.md` — the single deploy skill covering CLI install/auth, site linking, Git-based continuous deployment (the primary path), manual/local and CI deploys, and deploy troubleshooting. (Environment variables live in `netlify-config`; local dev lives in `netlify-frameworks`.)
 
+**Setting up A/B testing or a private beta by splitting traffic between deployed branches?**
+Read `netlify-split-testing/SKILL.md` for the branch-based Split Testing feature, the `nf_ab` cookie, client-side analytics tracking, and its conflicts with Edge Functions, Agent Runners, and Functions/proxy-generated content.
+
 **Want to run AI agents (Claude, Codex, Gemini) remotely on your site?**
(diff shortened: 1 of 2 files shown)


agroofficial1256-spec:
Netlify Split Testing is required-

Performance-Friendly: A/B Testing: lets you test two different website designs/web apps/web games  (like a new layout vs. an old one) to see which one performs better. Unlike traditional A/B testing tools that use JavaScript and slow down the page (causing a visible "flicker"), Netlify handles the split instantly at the server/CDN level. The site stays lightning fast.

Safe Beta Releases: It allows developers to safely roll out a new, experimental version of a site to a small, specific group of users (like an invite-only beta) before releasing it to the public.

Traffic Routing: It uses an nf_ab cookie to ensure a user stays on the same branch they were initially served. You can hijack this cookie manually for opt-in private betas.

The Core Benefit: It performs A/B testing at the edge (CDN level) using Git branches. Because it doesn't rely on client-side JavaScript to route traffic, there is no performance penalty or layout shift.

anish12345678910aa-coder (approved):

jaredm563:
Nice addition Anish! The skill is accurate and closely matches the official docs, including the limitations and the `nf_ab `opt in beta pattern (100% prod / 0% beta + client-side cookie). 

Two things before merge:

1. Blocking issue: the Hugo example is malformed. This line sits outside any code fence, right after the JS block closes, so it renders as loose body text instead of a code sample:

{{ getenv "BRANCH" }}  {# Hugo #}

I think it should wrap it in its own fenced block.

Example:

<pre> ``` {{ getenv "BRANCH" }} {# Hugo #} ``` </pre>

2. Suggestion: Note the Edge Functions / Agent Runners conflict higher. 

Right now the benefits lead and the "enabling Split Testing disables Edge Functions and breaks Agent Runners" tradeoff is kind of buried in the limitations list and recap. 

Since we push edge/agent primitives so heavily, thats probably the thing thats most likely to bite a user mid task. So, an agent reading this should hit it early so it doesnt recommend Split Testing to someone on an SSR/edge site. WDYT?

Maybe you could add a one line warning to the top? ex:

"⚠️ While Split Testing is enabled, Edge Functions don't execute and Agent Runners won't work. Don't recommend it for sites relying on edge/SSR/agent features."

Otherwise LGTM!

agroofficial1256-spec:
> Nice addition Anish! The skill is accurate and closely matches the official docs, including the limitations and the `nf_ab `opt in beta pattern (100% prod / 0% beta + client-side cookie).
> 
> Two things before merge:
> 
> 1. Blocking issue: the Hugo example is malformed. This line sits outside any code fence, right after the JS block closes, so it renders as loose body text instead of a code sample:
> 
> {{ getenv "BRANCH" }} {# Hugo #}
> 
> I think it should wrap it in its own fenced block.
> 
> Example:
> 
>  ``` {{ getenv "BRANCH" }} {# Hugo #} ``` 
> 2. Suggestion: Note the Edge Functions / Agent Runners conflict higher.
> 
> Right now the benefits lead and the "enabling Split Testing disables Edge Functions and breaks Agent Runners" tradeoff is kind of buried in the limitations list and recap.
> 
> Since we push edge/agent primitives so heavily, thats probably the thing thats most likely to bite a user mid task. So, an agent reading this should hit it early so it doesnt recommend Split Testing to someone on an SSR/edge site. WDYT?
> 
> Maybe you could add a one line warning to the top? ex:
> 
> "⚠️ While Split Testing is enabled, Edge Functions don't execute and Agent Runners won't work. Don't recommend it for sites relying on edge/SSR/agent features."
> 
> Otherwise LGTM!





--------commits done

seancdavis:
@agroofficial1256-spec Thanks for the PR! 💥 

We're actively migrating our skill generation to an automated system and aren't merging new skills at this time. We'll take this need into consideration for that system as we finish porting all of the existing skills.

I will close this, since the source content is expected to be sourced externally.

seancdavis:
For others who come looking for a split testing skill, you can use the content within this PR to create a local skill, or you can point your agent to [our docs page](https://docs.netlify.com/manage/monitoring/split-testing/).

PR #110, opened 2026-08-14
fix(ctx-receive): reject out-of-order deliveries

domitriusclark:
Monotonicity guard for the receiver: an out-of-order delivery (an older docs dispatch committing after a newer import) is skipped; anything unvalidatable fails closed.

Ordering is read from state.json's top-level `lastImportedCommit` — written by `ctx-receive.mjs` on every run (the writer was moved into this PR from #115 per its review, so the key ships with its consumer), with per-grouping `docsCommit` entries left as provenance that legitimately disagrees. The guard baselines against the unmerged rolling branch when it exists and treats a state predating the ordering key as bootstrap. An ordering-only advance (`state_changed` with no skill changes) is still committed so the baseline can't go stale on byte-identical redeliveries.

Guard logic lives in `scripts/ctx-ordering-guard.mjs` (13-case suite) and the ordering writer in `scripts/ctx-receive.mjs` (6-case suite); both run via `npm test` and the `test-receiver` job in validate-skills. Manual recovery via the `skip_guard` workflow_dispatch input.

Note for merge sequencing with #115: both touch `ctx-receive.mjs` and `package.json`'s test script — whichever lands second rebases (this branch's `node --test scripts/*.test.mjs` glob wins), and the validate-skills `paths:`/`test-receiver` overlap must be unioned, not chosen between.

## Summary by CodeRabbit

* **New Features**
  * Added safeguards to process documentation updates in the correct commit order.
  * Added manual recovery for skipped processing.
  * Import state now tracks the latest processed documentation commit.
  * Import detection now recognizes file, path, and executable-permission changes.

* **Bug Fixes**
  * Stale, duplicate, unrelated, or unverifiable deliveries are safely skipped.
  * Improved handling of missing, malformed, or invalid state.
  * Import and pull request creation now proceed only after ordering validation.

* **Tests**
  * Added automated coverage for ordering, recovery, and state scenarios.
  * Expanded workflow validation for receiver and ordering checks.


.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,56 @@ jobs:
...
+          last=$(node -e '
+            const s = require("./.ctx-gen/state.json");
+            const c = [...new Set(Object.values(s).map(v => v && v.docsCommit).filter(Boolean))];
+            if (c.length === 0) { console.log("BOOTSTRAP"); process.exit(0); }
+            if (c.length > 1) { console.error("state.json records conflicting docsCommit values: " + c.join(", ")); process.exit(1); }
+            if (!/^[0-9a-f]{40}$/.test(c[0])) { console.error("recorded docsCommit is not a commit hash: " + JSON.stringify(c[0])); process.exit(1); }
+            console.log(c[0]);
+          ') || { echo "::error::import state exists but cannot be validated — failing closed (fix or explicitly reset .ctx-gen/state.json)"; exit 1; }
+          [ "$last" = "BOOTSTRAP" ] && { echo "state has no recorded docsCommit — bootstrap import proceeds"; exit 0; }
+          incoming=$(git -C .docs-src rev-parse HEAD)
```
seancdavis: This reads `state.json` from `main`, but the last import usually isn't on main yet. It's sitting on the unmerged rolling branch. The PR step below does `checkout -B` from main and force-pushes, so main only learns about an import when a human merges.

Walk the race this guard exists for: docs dispatches Y, we import it, rolling PR opens. A few minutes later the older X lands (its AXIS gate just took longer). Guard reads main → still `e33a726` → X descends from it → passes. Import runs, force-push, and Y's content is gone. Silent rollback.

It only catches the case where someone merged between the two deliveries, which is the least likely timing. I'd fetch `origin/ctx-pipeline/agent-context-sync` and, when it exists, use *its* `.ctx-gen/state.json` as the baseline, falling back to main's.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
...
+          last=$(node -e '
+            const s = require("./.ctx-gen/state.json");
+            const entries = Object.values(s);
+            if (entries.length === 0) { console.error("state.json is present but empty — not a bootstrap; reset it explicitly if intended"); process.exit(1); }
+            // EVERY entry must carry the same valid commit — a partial state
+            // (some entries missing docsCommit) is corruption, not a quorum.
+            const commits = entries.map(v => v && v.docsCommit);
+            if (commits.some(c => !/^[0-9a-f]{40}$/.test(c || ""))) { console.error("state entry missing or malformed docsCommit — state is partial/corrupt"); process.exit(1); }
+            const distinct = [...new Set(commits)];
+            if (distinct.length > 1) { console.error("state.json records conflicting docsCommit values: " + distinct.join(", ")); process.exit(1); }
```
seancdavis: This is the one that worries me most. `ctx-receive.mjs` only rewrites `state[grouping]` for groupings whose `source_hash` changed (lines 137-142), so `docsCommit` drifts apart across groupings by design. The first dispatch that touches only `functions` leaves `functions` at X and the other twelve at `e33a726`. Rolling PR merges, the next dispatch hits `distinct.length > 1`, and every run fails until someone hand-edits `state.json`.

Today all 13 agree only because the bootstrap imported everything in one go.

I think CodeRabbit's original read was right — mixed commits are normal here. The check should be a loop: for each distinct commit, `is-ancestor $c $incoming`; any `1` means skip. Not fail.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
...
+
+
 
       - name: Resolve docs commit
         id: docs
         run: echo "sha=$(git -C .docs-src rev-parse HEAD)" >> "$GITHUB_OUTPUT"
 
       - name: Import changed skills
         id: import
+        if: env.SKIP_IMPORT != '1'
```
seancdavis: Skipping here leaks into the PR step. Its condition is `success() && steps.import.outputs.changed_count != '0'`. When this step is skipped, `changed_count` is empty, `'' != '0'` is true, and `success()` is still true (skipped isn't failed). So the PR step runs, `git commit` has nothing staged, exits 1 under `bash -e`, and the job goes red. Every stale delivery becomes a failure instead of a quiet no-op.

Needs `env.SKIP_IMPORT != '1'` (or `steps.import.outcome == 'success'`) on that step too.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
...
+            const entries = Object.values(s);
+            if (entries.length === 0) { console.error("state.json is present but empty — not a bootstrap; reset it explicitly if intended"); process.exit(1); }
+            // EVERY entry must carry the same valid commit — a partial state
+            // (some entries missing docsCommit) is corruption, not a quorum.
+            const commits = entries.map(v => v && v.docsCommit);
+            if (commits.some(c => !/^[0-9a-f]{40}$/.test(c || ""))) { console.error("state entry missing or malformed docsCommit — state is partial/corrupt"); process.exit(1); }
+            const distinct = [...new Set(commits)];
+            if (distinct.length > 1) { console.error("state.json records conflicting docsCommit values: " + distinct.join(", ")); process.exit(1); }
+            console.log(distinct[0]);
+          ') || { echo "::error::import state exists but cannot be validated — failing closed (fix or explicitly reset .ctx-gen/state.json)"; exit 1; }
```
seancdavis: A few of these say "fix or explicitly reset `state.json`," but there's no way to do that from the workflow. `workflow_dispatch` has no bypass input, so recovery is a hand-authored commit to main. A `skip_guard` boolean input, honored only on manual runs, would cover it.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
...
+          # Exactly one distinct recorded docsCommit, and it must look like a
+          # commit hash. Disagreement or garbage = corrupt state: fail, don't guess.
+          last=$(node -e '
+            const s = require("./.ctx-gen/state.json");
+            const entries = Object.values(s);
+            if (entries.length === 0) { console.error("state.json is present but empty — not a bootstrap; reset it explicitly if intended"); process.exit(1); }
+            // EVERY entry must carry the same valid commit — a partial state
+            // (some entries missing docsCommit) is corruption, not a quorum.
+            const commits = entries.map(v => v && v.docsCommit);
+            if (commits.some(c => !/^[0-9a-f]{40}$/.test(c || ""))) { console.error("state entry missing or malformed docsCommit — state is partial/corrupt"); process.exit(1); }
```
seancdavis: Small one: the script writes `docsCommit: opts.docsCommit || manifest.generated_from?.commit || null` (`ctx-receive.mjs:139`). CI always passes `--docs-commit`, but one local run without it writes `null` or a short SHA into state, and this regex then fails every future CI run. I'd have the script require a full SHA rather than trusting it here.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
...
+          ') || { echo "::error::import state exists but cannot be validated — failing closed (fix or explicitly reset .ctx-gen/state.json)"; exit 1; }
+          incoming=$(git -C .docs-src rev-parse HEAD)
+          if ! git -C .docs-src cat-file -e "$last" 2>/dev/null; then
+            echo "::error::last imported commit $last is not present in the docs checkout — cannot establish ordering (incomplete fetch or rewritten history); failing closed"
+            exit 1
+          fi
+          rc=0
+          git -C .docs-src merge-base --is-ancestor "$last" "$incoming" || rc=$?
+          if [ "$rc" = 1 ]; then
+            echo "::notice::stale delivery: incoming $incoming does not descend from last imported $last — skipping (a newer import already covers this)"
```
seancdavis: `is-ancestor` returning 1 also means "unrelated," not just "older." A manual run pointed at a docs feature branch lands here too, and the log would claim a newer import covers it. Maybe just "does not descend from last imported" and leave it there.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
           token: ${{ secrets.DOCS_READ_TOKEN }}
           path: .docs-src
           persist-credentials: false
+          # Full history: the monotonicity guard below needs ancestry.
+          fetch-depth: 0
```
seancdavis, line 101: This clones all of netlify/docs on every dispatch to answer one ancestry question. `git fetch --shallow-exclude="$last"` (or fetching `$last` and deepening) gets the same answer a lot cheaper. Not blocking, but this runs on every docs regen.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
...
+          if [ ! -e .ctx-gen/state.json ] && [ ! -L .ctx-gen/state.json ]; then
+            echo "no prior import state — bootstrap import proceeds"
+            exit 0
+          fi
+          # Present but not a regular file (dir/FIFO/symlink): fail before
+          # node — require() on a dir can execute code, on a FIFO can hang.
+          if [ ! -f .ctx-gen/state.json ] || [ -L .ctx-gen/state.json ]; then
+            echo "::error::state.json exists but is not a regular file — failing closed"
+            exit 1
+          fi
```
seancdavis: I'd drop the dir/FIFO/symlink dance. Anyone who can turn `state.json` into a directory can also edit this workflow, so it doesn't buy anything, and it makes the step harder to read than the actual logic deserves.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
...
+            if (entries.length === 0) { console.error("state.json is present but empty — not a bootstrap; reset it explicitly if intended"); process.exit(1); }
+            // EVERY entry must carry the same valid commit — a partial state
+            // (some entries missing docsCommit) is corruption, not a quorum.
+            const commits = entries.map(v => v && v.docsCommit);
+            if (commits.some(c => !/^[0-9a-f]{40}$/.test(c || ""))) { console.error("state entry missing or malformed docsCommit — state is partial/corrupt"); process.exit(1); }
+            const distinct = [...new Set(commits)];
+            if (distinct.length > 1) { console.error("state.json records conflicting docsCommit values: " + distinct.join(", ")); process.exit(1); }
+            console.log(distinct[0]);
+          ') || { echo "::error::import state exists but cannot be validated — failing closed (fix or explicitly reset .ctx-gen/state.json)"; exit 1; }
+          incoming=$(git -C .docs-src rev-parse HEAD)
```
seancdavis: Nit: `incoming` is the same thing "Resolve docs commit" computes two steps down. This step could emit it as an output and drop the other. Also two trailing blank lines at the end of this block.

seancdavis (changes requested):
Thanks for picking this up. Out-of-order deliveries are a real hole, and ancestry is the right tool for it. But I walked through how `ctx-receive.mjs` writes `state.json`, and the guard as written breaks the pipeline instead of protecting it. Requesting changes. Three things block, each with an inline comment:

1. `state.json` isn't supposed to have one `docsCommit`. The script only bumps `state[grouping]` for groupings whose content changed, so the first ordinary partial import leaves groupings on different commits, and the "conflicting docsCommit" check fails every run after that. Forever. It should be "incoming descends from each recorded commit," not "all recorded commits agree."

2. The guard reads `main`'s state, but the last import lives on the unmerged rolling branch. Two dispatches minutes apart (the case this is for) both see the same stale baseline on main, and the older one still force-pushes over the newer one. We need to read the rolling branch's `state.json` when it exists.

3. When it does skip, the run still goes red, because the PR step's `changed_count != '0'` is true for an empty output. Gate that step on the skip too.

Smaller stuff inline: no bypass input for manual recovery, the regex vs the script's `null`/short-SHA fallback, a misleading "newer import covers this" notice, a full clone of docs every run, and I'd cut the symlink/FIFO checks.

One ask before the next round: run the node snippet against a `state.json` where two groupings have different commits. That reproduces #1 in about ten seconds. And nothing in CI exercises this workflow, so I'd want that kind of check on the rest of it too.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
...
+          fi
+          # Present but not a regular file (dir/FIFO/symlink): fail before
+          # node — require() on a dir can execute code, on a FIFO can hang.
+          if [ ! -f .ctx-gen/state.json ] || [ -L .ctx-gen/state.json ]; then
+            echo "::error::state.json exists but is not a regular file — failing closed"
+            exit 1
+          fi
+          # Exactly one distinct recorded docsCommit, and it must look like a
+          # commit hash. Disagreement or garbage = corrupt state: fail, don't guess.
+          last=$(node -e '
```
seancdavis: The guard is right; it's reading the wrong field. This block conflicts with #115, which is open now.

#115 makes import decisions on byte difference rather than manifest hash, and as a deliberate consequence a grouping whose output is byte-identical **writes no state entry**. Its header comment calls this out:

> a regeneration whose output is byte-identical to what's already imported imports nothing and writes no state entry, so `state.json` provenance can lag the newest `source_hash`. Harmless

So after #115 the normal steady state is: grouping A imported at commit X, grouping B untouched and still recording W. Two distinct `docsCommit` values — which line 129 treats as corruption and fails the run on. Whichever of these merges second, the receive workflow starts failing closed on ordinary runs.

Underneath that is a design question worth settling before either lands: **is `state.json` a provenance log or the ordering authority?** It can't be both. Right now #115 treats it as a log that may lag, and this guard treats it as the source of truth for "what did we last import."

My read is that we should split them rather than pick a winner, because deriving "where am I" by aggregating "what happened" is the thing this class of system consistently avoids — Terraform bumps `serial` on every apply including no-ops while the resource map only tracks real resources; Kafka commits a consumer offset even when a poll produces nothing downstream; git writes `HEAD` on every operation and never recomputes it by scanning commits.

Concretely, I'd change #115 to write a top-level `lastImportedCommit` unconditionally on every run — no-op included — and leave the per-grouping entries as provenance. Then this guard gets simpler:

- read `state.lastImportedCommit` instead of aggregating `Object.values(s)`
- drop the every-entry-must-have-a-valid-`docsCommit` requirement (lines 126–127)
- drop the `distinct.length > 1` conflict check (lines 128–129) — under #115 that's the expected state, not corruption
- treat a present `state.json` with no ordering key as **bootstrap**, not corruption, so this doesn't fail closed on every repo until #115 lands

That takes this from ~12 lines of validation down to reading one field.

Everything else here I'd keep exactly as-is: the fail-closed posture, the not-a-regular-file check on lines 113–117, `merge-base --is-ancestor`, and `fetch-depth: 0`. The reasoning in the comments is good and the failure modes you're guarding are real ones.

Happy to make the #115 side of this change so they land together — #115 needs to go first either way, since this guard depends on the key existing.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,13 +92,67 @@ jobs:
...
+            const s = require("./.ctx-gen/state.json");
+            const entries = Object.values(s);
+            if (entries.length === 0) { console.error("state.json is present but empty — not a bootstrap; reset it explicitly if intended"); process.exit(1); }
```
seancdavis: Once the guard reads a single top-level key, this empty-state branch can fold into the bootstrap case above — an empty object and a missing file both mean "nothing imported yet."

Worth keeping the distinction only if you want an explicit reset to be loud rather than silent. If so, the message could say which of the two it hit.

seancdavis (changes requested):
These are updates after addressing design conflicts from #115

domitriusclark:
Reworked in 3a52854 per the review — taking the split you proposed (and that #115 now implements with `lastImportedCommit`):

**The three blockers**

1. **Mixed per-grouping commits no longer matter.** The guard reads only the top-level `lastImportedCommit` ordering key from #115. Per-grouping `docsCommit` entries are never consulted — they're provenance and are expected to disagree. A state with no ordering key (or an empty object) is treated as bootstrap, not corruption, so this doesn't fail closed on every repo until #115 lands; the guard simply becomes effective on the first import after #115. Ran your repro: a state with `functions` and `platform` on different commits proceeds as bootstrap — it's now a test case (`bootstrap: state without an ordering key proceeds`).
2. **Baseline comes from the rolling branch.** A new "Resolve import baseline" step fetches `origin/ctx-pipeline/agent-context-sync` and reads its `.ctx-gen/state.json` when the branch exists, falling back to main's. The two-dispatches-minutes-apart race now sees Y's position on the unmerged branch.
3. **A skipped delivery is a quiet no-op.** The PR step is gated on the guard's `skip` output as well as `changed_count`, so the empty-output `'' != '0'` path can't reach `git commit`.

**The smaller items**

- `skip_guard` boolean `workflow_dispatch` input for manual recovery, honored only on manual runs (logged as a workflow warning when used).
- Skip message now says only what's known: "incoming X does not descend from last imported Y" — no "newer import covers this" inference (unrelated histories land in the same branch).
- The regex/short-SHA concern: the guard fails closed on a non-40-hex ordering key, recoverable via `skip_guard`. Script-side full-SHA validation of `--docs-commit` belongs in `ctx-receive.mjs`, which #115 rewrites — deferred to avoid colliding; CI always passes `rev-parse HEAD` so it can't hit this path.
- Symlink/FIFO dance is gone, but stayed fail-closed: baseline state is read as a git blob via `git show` and `JSON.parse`'d in a child process — never `require()`'d from the working tree. A symlink blob is its target text and a directory a tree listing; both fail JSON parsing.
- `incoming` is resolved once in "Resolve docs commit" and passed in; trailing blank lines gone. Kept `fetch-depth: 0` per your follow-up.

**CI now exercises this**

The guard logic moved to `scripts/ctx-ordering-guard.mjs` (zero-dep, same style as `ctx-receive.mjs`) with a 12-case suite in `scripts/ctx-ordering-guard.test.mjs` covering the full decision table against real temp git repos: three bootstrap shapes, descent/same-commit proceed, older/unrelated skip, and five fail-closed cases. Wired as `npm test` (`node --test scripts/*.test.mjs`, so it'll pick up #115's suite too) and a `test-receiver` job in `validate-skills.yml` matching the shape #115 adds — whichever lands second resolves a trivial union.

**Forward-compat with #115's ordering-only advances:** the PR step also commits when `state_changed == 'true'` with `changed_count == 0` (labeled "none (import ordering advanced)"), so a byte-identical redelivery still advances the guard's baseline instead of leaving it stale. That output doesn't exist pre-#115, where the condition degrades to exactly today's behavior.

Agreed #115 should merge first; this is written to be safe in either order.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,26 +97,91 @@ jobs:
...
+            echo "::warning::monotonicity guard bypassed via skip_guard (manual run) — importing without an ordering check"
+            echo "skip=0" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          if [ -z "$STATE" ]; then
+            echo "no prior import state — bootstrap import proceeds"
+            echo "skip=0" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          node scripts/ctx-ordering-guard.mjs --docs .docs-src --incoming "$INCOMING" --state "$STATE"
```
seancdavis, line 159: **Blocking** — this should not merge before #115.

`lastImportedCommit` doesn't exist yet: `ctx-receive.mjs` on this branch writes neither it nor `state_changed`. Merged alone, the guard reads no position, treats every delivery as bootstrap, and waves everything through. Harmless, but it means the protection silently isn't there — and the ordering-only advance the PR step now depends on never fires.

.github/workflows/validate-skills.yml
```diff
@@ -56,3 +58,16 @@ jobs:
 
           run_with_retries skill-validator check --strict --emit-annotations skills/
           run_with_retries skill-validator check --strict -o markdown skills/ >> "$GITHUB_STEP_SUMMARY"
+
+  test-receiver:
```
seancdavis, line 66: **Follow-up** — #115 adds a job with this same name at this same position, and adds its own two entries to the `paths:` list above. Whichever lands second conflicts.

The risk is in the resolution, not the conflict. Take one side's block wholesale and the other side's `paths:` entries disappear — after which a PR touching only `scripts/ctx-receive.mjs`, the file that writes the key this guard reads, triggers no workflow and runs no tests. Nothing goes red; the job just never fires. The two lists need to be unioned, not chosen between.

scripts/ctx-ordering-guard.mjs
```diff
@@ -0,0 +1,127 @@
...
+
+  if (git(opts.docs, ['cat-file', '-e', `${last}^{commit}`]) !== 0) {
+    fail(`last imported commit ${last} is not present in the docs checkout — cannot establish ordering (incomplete fetch or rewritten history); failing closed`);
+  }
+
+  const rc = git(opts.docs, ['merge-base', '--is-ancestor', last, opts.incoming]);
+  if (rc === 0) {
+    emit(false, `incoming ${opts.incoming} descends from last imported ${last} — proceeding`);
+  } else if (rc === 1) {
+    emit(true, `incoming ${opts.incoming} does not descend from last imported ${last} — skipping this delivery`);
```
seancdavis, line 121: **Follow-up** — a skip is a plain log line and a green run: no `::warning::`, no step summary, no annotation. The bypass path one screen up gets a warning, so the more dangerous action is the more visible one.

Most skips are self-healing, since the receiver imports the full tree rather than a delta — a later docs commit carries the skipped content anyway. The case that doesn't heal is a `lastImportedCommit` that main no longer descends from (a docs history rewrite): the commit still resolves, so this doesn't fail closed, and every future delivery skips green forever. Tracked in AX-159; not asking for it here.

scripts/ctx-ordering-guard.test.mjs
```diff
@@ -0,0 +1,167 @@
...
+  assert.equal(r.output, '');
+});
+
+test('fail closed: state is not valid JSON', () => {
+  const r = runGuard({ state: 'not json{', incoming: docs.b });
+  assert.equal(r.status, 1);
+  assert.match(r.stderr, /not valid JSON/);
+});
+
+test('fail closed: state is a JSON array, not an object', () => {
```
seancdavis, line 173: **Consider** — the `merge-base` error branch (status other than 0 or 1) has no test, and it's reachable: `cat-file -e` only validates the recorded commit, so an unresolvable `--incoming` sails past it and makes `merge-base` exit 128.

That exact silent-bypass is what commit `8a32311` was fixing, and nothing pins it now.

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,26 +97,91 @@ jobs:
           token: ${{ secrets.DOCS_READ_TOKEN }}
           path: .docs-src
           persist-credentials: false
+          # Full history: the monotonicity guard below needs ancestry.
+          fetch-depth: 0
```
seancdavis, line 101: **Consider** — full history of netlify/docs on every dispatch. Ancestry needs it, so this looks like the right call rather than a mistake; flagging it as a per-delivery cost worth having consciously accepted on a repo that size.

seancdavis (commented):
Commenting for now so that you can address while I work on cleaning up #115, which needs to come first.

seancdavis:
Moved in from #115 (per its review): the writer for `lastImportedCommit` and the `state_changed` output now lives here with its consumer. `16be109` ports it onto this branch's receiver (the `source_hash` skip is untouched) with `scripts/ctx-receive-ordering.test.mjs` (6 cases: first import, same/newer-commit re-dispatch, no `--docs-commit`, `--dry-run`, reserved-name collision); `4091871` adds the receiver paths to the CI filter. #115 reverted its copy.

After #115 merges this branch needs a rebase — both touch `ctx-receive.mjs`, and `package.json`'s test script conflicts (this branch's `node --test scripts/*.test.mjs` glob should win).

scripts/ctx-ordering-guard.test.mjs
domitriusclark: Pinned in 0ceb9b3: a resolvable recorded commit + unresolvable `--incoming` now asserts exit 1 on the `merge-base failed with status` branch with no GITHUB_OUTPUT written. Good catch — that was exactly the 8a32311 shape with nothing holding it.

.github/workflows/validate-skills.yml
domitriusclark: Agreed — noting for whoever resolves it: the `paths:` lists and the `test-receiver` block both need the union, not either side. Since #115 reverted its receiver/ordering pieces into this branch (16be109/4091871 already carry the receiver paths here), the remaining overlap should be small; will verify the union explicitly on the post-#115 rebase.

.github/workflows/ctx-pipeline-receive.yml
domitriusclark: Consciously accepted: ancestry needs the history, dispatches are regen-frequency (not commit-frequency), and a shallow-exclude scheme would need its own fail-closed handling for "commit not in the shallow window" — more machinery than the cost saves. Can revisit if docs clone time ever becomes the long pole in this job.

.github/workflows/ctx-pipeline-receive.yml
domitriusclark: Resolved by your 16be109/4091871 — the writer now ships with its consumer, so the guard is protective standalone rather than inert-until-#115. PR body updated to reflect the new coupling (rebase + union notes included).

seancdavis:
Brought #115 in now rather than rebasing after it merges: `eceaaa9` merges its branch (head `6a4b650`) with a merge commit, no history rewritten. Resolution: `ctx-receive.mjs` keeps #115's byte-diff + `intermediateHash` mutations and this PR's `lastImportedCommit`, with one hoisted state write after both and `state_changed` emitted; `package.json` keeps the `node --test scripts/*.test.mjs` glob; CI path filters unioned, Node 18 setup and `contents: read` retained. The "State advanced" log now fires only when the ordering key actually moved, so a drift-warning run at the same commit doesn't claim an advance. #115's first-import test gained the `lastImportedCommit` it now expects.

Base is temporarily #115's branch so this diff shows only its own changes; GitHub retargets to `main` when #115 squashes and its branch is deleted. Suite: 41/41 across all three test files (also merged in `0ceb9b3` from origin).

.github/workflows/ctx-pipeline-receive.yml
```diff
@@ -92,26 +97,91 @@ jobs:
...
+      # the PR step below force-pushes ctx-pipeline/agent-context-sync and main
+      # only learns of an import when a human merges. Guarding against main's
+      # copy alone would let two dispatches minutes apart both see the same
+      # stale baseline — exactly the race the guard exists for.
+      - name: Resolve import baseline
+        id: baseline
+        run: |
+          branch="ctx-pipeline/agent-context-sync"
+          ref="HEAD"
+          if git fetch -q origin "$branch" 2>/dev/null; then
```
seancdavis, line 117: **Follow-up** — every `git fetch` failure lands in the same `else`, so a transient one silently downgrades the baseline to main's older `state.json` and logs "no rolling branch" as though the branch weren't there.

That's the one fail-open in an otherwise fail-closed design, and it opens exactly when the guard matters: main's state is stale by definition between imports, so a stale delivery arriving during a fetch blip passes the ancestry check against a position two imports old. `2>/dev/null` drops the reason too, so the run log can't tell you afterwards which case you were in.

scripts/ctx-receive.mjs
```diff
@@ -289,24 +315,38 @@ function main() {
...
-    const current = fs.existsSync(opts.state) ? fs.readFileSync(opts.state, 'utf8') : null;
-    const stateChanged = current !== nextState;
-    if (stateChanged) fs.writeFileSync(opts.state, nextState);
-  }
+  // The ordering key advances on every run, imports or not — that's what makes
+  // it usable for ordering. Only write when we were actually told which commit
+  // we're importing; inventing one from a manifest would record a position we
+  // can't defend.
+  const prevOrdering = state[ORDERING_KEY];
+  if (!opts.dryRun && opts.docsCommit) state[ORDERING_KEY] = opts.docsCommit;
```
seancdavis, line 323: **Consider** — `--docs-commit` goes into the ordering key with no shape check, and the guard's `/^[0-9a-f]{40}$/` is the only thing that ever looks at it.

I raised this last round and the thread didn't get an answer, so re-flagging rather than assuming: one local run with a short SHA writes a value that fails closed on every subsequent CI run, and the recovery is knowing `skip_guard` exists. Reasonable to leave the writer dumb on purpose — the header argues for exactly that — but if that's the call I'd rather it be written down than inferred.

Related, smaller: the PR body says the key is written "on every run," and it's every run that supplies `--docs-commit`. The test at `ctx-receive-ordering.test.mjs:149` pins the exception, so it's the description that's off, not the code.

.github/workflows/validate-skills.yml
```diff
@@ -61,8 +64,10 @@ jobs:
           run_with_retries skill-validator check --strict -o markdown skills/ >> "$GITHUB_STEP_SUMMARY"
 
   test-receiver:
-    # Zero-dependency script tests for the Stage 2 receiver (scripts/ctx-receive.mjs).
-    # No npm install: the package has no dependencies, so `npm test` runs as-is.
+    # Zero-dependency script tests for the Stage 2 receiver pipeline
+    # (scripts/ctx-receive.mjs, scripts/ctx-ordering-guard.mjs); `npm test`
+    # runs every scripts/*.test.mjs. No npm install: the package has no
```
seancdavis, line 69: **Consider** — 60-odd tests now cover both scripts, and still nothing covers the workflow that calls them.

The part that decides whether a skip is a quiet no-op or a red run is the `if:` chain in `ctx-pipeline-receive.yml` and the shell in `Resolve import baseline` — the implicit `success()` on the import step, `state_changed == 'true'`, the `git show` blob read. That reasoning lives in comments and in our heads. I raised the same gap in the first round; the script suites since then are real progress, but they stop at the boundary where the bugs from round one actually were.

seancdavis (approved):
A couple of thoughts and follow ups for consider. OTherwise all good.

Claude summary: 

---

Re-reviewed against the merged tree. I resolved the conflicts with main in `ddbd931` — four files, all the same shape (this branch already carried #115's work, main got it via `bacaa0d`), so every hunk was branch-side-newer. I kept the branch side in each and checked it was a superset rather than a coin flip: the `validate-skills` `paths:` list is the union of both sides, the `test-receiver` job body was byte-identical on both, and `package.json`'s `node --test scripts/*.test.mjs` covers all three suites. That's the union Dom and I flagged, done deliberately. CI is green on it.

The blocking condition from last round is gone. `ctx-receive.mjs` writes `lastImportedCommit` here now, so the writer ships with its consumer instead of the guard being inert until #115 landed. Splitting the ordering key from the per-grouping provenance was the right call, and the guard reading one field instead of reconciling thirteen is what makes the rest of this readable.

The tests earn their keep. `ctx-receive-ordering.test.mjs` asserts the exact `GITHUB_OUTPUT` string rather than probing individual fields, so a wrong implementation fails rather than passes differently, and `0ceb9b3` pinned the `merge-base` error branch I asked about.

**Recommended verdict: approve.** One follow-up and two considers below, none of them holding the merge.

Housekeeping: the PR body's closing paragraph still describes #115 as unlanded and the rebase as pending. Worth a trim before merge so the next reader isn't looking for work that's done. The skip-visibility gap from last round is still tracked in [AX-159](https://linear.app/netlify/issue/AX-159/stage-2-ordering-guard-a-skipped-delivery-is-invisible-and-a-stuck) and unchanged here.

PR #115, opened 2026-08-20
fix: import agent-context skills on byte difference, not manifest hash

seancdavis:
## Summary

The Stage-2 receiver now decides import vs skip by byte-comparing `agent-context/<grouping>/skill/**` against `skills/<name>/**`, so upstream hand edits (like netlify/docs#801) propagate instead of being silently skipped when the manifest's `source_hash` hasn't moved. `changed_count > 0` now always implies a real git diff, which also removes the empty-commit failure mode in the receive workflow.

## What changed

- `scripts/ctx-receive.mjs`: `treeDiffers()` (path set + file bytes + executable bit) is the sole import/skip decision; manifest state is provenance-only
- Symlinks and other non-regular entries in either tree fail the run loudly (`cpSync` would copy them, the gate can't compare them); the source tree is validated before the first-import short-circuit
- A destination that exists but isn't a directory counts as changed and is replaced
- A grouping with no `skill/SKILL.md` is skipped only if never imported; once imported, its disappearance fails the run rather than leaving a stale skill
- `context.md` + `system.md` are hashed per grouping into `state.json` (`intermediateHash`); when that moves while `skill/` is byte-identical the run **warns** (`[warn]` + `::warning::` on Actions), once per upstream change, and imports nothing — never fails. Warn-once persistence becomes effective when #110's `state_changed` commit gate lands
- `importerVersion` removed (script, `config.json`, `state.json`, README): with a faithful byte copy no version bump can change output
- `scripts/ctx-receive.test.mjs` (new): 22-case zero-dep suite, one fixture per case; wired as `npm test`
- `.github/workflows/validate-skills.yml`: `test-receiver` job (Node 18 via SHA-pinned `setup-node`, `contents: read`); PR path filter covers the receiver scripts and `package.json`
- `.ctx-gen/README.md`: `state.json` described as a provenance log; the delta is the byte comparison
- `docs/autopilot/`: run spec + report with both iteration rounds ([full audit detail](docs/autopilot/2026-08-19-ax-136-receiver-byte-diff-report.md))

Ordering (`lastImportedCommit`, `state_changed`, the monotonicity guard) lives in #110, which should merge after this and needs a rebase.

## Verify

- After netlify/docs#801 merges, the next receive run's rolling sync PR should contain the Next.js fetch-skew qualification — that's the live proof of this fix.

Linear: [AX-136](https://linear.app/netlify/issue/AX-136/stage-2-receiver-silently-skips-hand-edited-upstream-skills)

## Summary by CodeRabbit

* **New Features**
  * Imports now detect file-path, content, and executable-permission changes.
  * Invalid destination files are replaced, and missing directories are repaired.
  * Intermediate file changes are tracked, with warnings when generated skills remain unchanged.
  * Incomplete groupings without a `SKILL.md` are skipped only when never previously imported.
* **Bug Fixes**
  * Unsupported symlinks and other non-regular entries now fail clearly.
  * Import state remains accurate after source or intermediate changes.
* **Tests**
  * Expanded coverage for edge cases, dry runs, and failure handling.
* **Documentation**
  * Added receiver design and implementation reports.
* **Chores**
  * CI runs receiver tests when relevant scripts or package configuration changes.

### DIFF
--- a/.github/workflows/validate-skills.yml
+++ b/.github/workflows/validate-skills.yml
@@ -3,6 +3,8 @@ on:
   pull_request:
     paths:
       - "skills/**"
+      - "scripts/ctx-receive.mjs"
+      - "scripts/ctx-receive.test.mjs"
       - ".github/workflows/validate-skills.yml"
 
 jobs:
@@ -56,3 +58,15 @@ jobs:
 
           run_with_retries skill-validator check --strict --emit-annotations skills/
           run_with_retries skill-validator check --strict -o markdown skills/ >> "$GITHUB_STEP_SUMMARY"
+
+  test-receiver:
+    # Zero-dependency script tests for the Stage 2 receiver (scripts/ctx-receive.mjs).
+    # No npm install: the package has no dependencies, so `npm test` runs as-is.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
+        with:
+          persist-credentials: false
+
+      - name: Run receiver tests
+        run: npm test
--- a/.github/workflows/validate-skills.yml
+++ b/.github/workflows/validate-skills.yml
@@ -5,6 +5,7 @@ on:
       - "skills/**"
       - "scripts/ctx-receive.mjs"
       - "scripts/ctx-receive.test.mjs"
+      - "package.json"
       - ".github/workflows/validate-skills.yml"
 
 jobs:
@@ -67,6 +68,11 @@ jobs:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
         with:
           persist-credentials: false
+
+      # 18 is the floor the script header promises; running on it keeps that claim honest.
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 18
 
       - name: Run receiver tests
         run: npm test
--- a/.github/workflows/validate-skills.yml
+++ b/.github/workflows/validate-skills.yml
@@ -63,6 +63,8 @@ jobs:
   test-receiver:
     # Zero-dependency script tests for the Stage 2 receiver (scripts/ctx-receive.mjs).
     # No npm install: the package has no dependencies, so `npm test` runs as-is.
+    permissions:
+      contents: read
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
--- a/docs/autopilot/2026-08-19-ax-136-receiver-byte-diff-report.md
+++ b/docs/autopilot/2026-08-19-ax-136-receiver-byte-diff-report.md
@@ -1,0 +1,53 @@
+# AX-136 autopilot run report — 2026-08-20
+
+Spec: `2026-08-19-ax-136-receiver-byte-diff.md`. Branch built unattended
+(launched 3:02 AM); orchestrator judged, Sonnet developers wrote, Codex
+audited read-only.
+
+## Slices
+
+1. `577e427` — byte-diff delta (`treeDiffers`) replaces the manifest-hash
+   skip; import reasons now name the hand-edit shape. Developer also made the
+   missing-`SKILL.md` check unconditional (previously only on the changed
+   path) — judged in-scope: same failure surface, now deterministic.
+2. `3472da7` — zero-dep test suite (5 tests incl. the docs#801 hand-edit
+   shape), `npm test` wired.
+3. `0fe8cdc` — `test-receiver` job in `validate-skills.yml`; PR paths trigger
+   extended to the receiver scripts so the suite actually fires on
+   receiver-only PRs.
+
+## Spec amendment (orchestrator, mid-run)
+
+The original done-signal grep (`prev.sourceHash === sourceHash` → 0 hits) was
+too blunt: it caught a benign log-only comparison that labels the hand-edit
+case. Amended to target the decision conjunction
+(`… && prev.importerVersion` → 0 hits) plus `treeDiffers` ≥ 2. Rationale
+recorded in the spec's Done-signal section.
+
+## Audit rounds (bound: 3, used: 1)
+
+- **Security (Codex):** clean, round 1.
+- **Simplicity (Codex):** 3 low findings.
+  - Import-reason ternary — accepted narrowed: label made factual
+    (`surface differs, source_hash unchanged`) + legacy-state guard; the
+    labeling itself kept deliberately (visibility is AX-136's point). Codex's
+    claim that it "couples the decision to manifest state" was wrong — the
+    decision is `treeDiffers` only.
+  - Redundant `Set` in `treeDiffers` — accepted; elementwise sorted-array
+    compare.
+  - Fold `test-receiver` into `validate` job — **rejected**: separate job is
+    a named status check, runs parallel to the validator's network download,
+    keeps jobs single-purpose.
+- Fix commit `5bfff01`; tests re-run green by orchestrator; follow-up
+  simplicity pass clean.
+
+## Completeness gate
+
+- `npm test` → 5/5 pass (orchestrator-run).
+- `rg "prev.sourceHash === sourceHash && prev.importerVersion"` → 0 hits.
+- `rg -c "treeDiffers"` → 3 (≥ 2).
+- `rg -c "npm test" .github/workflows/validate-skills.yml` → 2 (≥ 1).
+- human-verify (for Sean): after netlify/docs#801 merges, the next receive
+  run's rolling sync PR must contain the Next.js fetch-skew qualification.
+
+No unmet items; no deferred findings beyond the rejected CI-job fold above.
--- a/docs/autopilot/2026-08-19-ax-136-receiver-byte-diff-report.md
+++ b/docs/autopilot/2026-08-19-ax-136-receiver-byte-diff-report.md
@@ -47,7 +47,37 @@ recorded in the spec's Done-signal section.
 - `rg "prev.sourceHash === sourceHash && prev.importerVersion"` → 0 hits.
 - `rg -c "treeDiffers"` → 3 (≥ 2).
 - `rg -c "npm test" .github/workflows/validate-skills.yml` → 2 (≥ 1).
-- human-verify (for Sean): after netlify/docs#801 merges, the next receive
-  run's rolling sync PR must contain the Next.js fetch-skew qualification.
+- human-verify (for Sean) — **pending, needs a live run**: after
+  netlify/docs#801 merges, the next receive run's rolling sync PR must contain
+  the Next.js fetch-skew qualification.
 
-No unmet items; no deferred findings beyond the rejected CI-job fold above.
+All machine-checkable items met; the human-verify item above is the one
+open item. No deferred findings beyond the rejected CI-job fold above.
+
+## Iteration 1 — 2026-08-25 (PR review)
+
+Input: domitriusclark's review on #115 (6 inline + 2 top-level) plus two
+CodeRabbit findings. All validated against the branch head `f3cae6a`.
+
+Dispositions:
+
+- Fixed — null `docsCommit` crash (test pins the manifest fallback);
+  missing `SKILL.md` warns and continues instead of failing the run; symlinks
+  and other non-regular entries fail loudly, executable bit compared (the
+  only mode bit git tracks); non-directory destination counts as changed and
+  is replaced; `importerVersion` removed from script, config, state, README
+  (with a faithful copy no bump can change output); tests isolated per
+  fixture; CI path filter includes `package.json`, `setup-node` pinned at
+  Node 18; README describes `state.json` as provenance. Commits `ace7e5f`,
+  `a701fb9`, `e22e3b4`, `1d9ae10`, `eb3f515`.
+- Moved — `lastImportedCommit` / `state_changed` had no consumer in this PR;
+  the consumer is #110 (`ctx-receiver-monotonic`). Reverted out of #115
+  (`1a0b93e`); the writer ships in #110 so that PR is self-contained.
+- Spec amended first (Program design): symlink/exec-bit policy,
+  `importerVersion` removal, warn-not-fail on missing `SKILL.md`.
+
+Audit (round 1 of 3): security clean; simplicity 1 low (test-helper state
+and unused returns) — accepted, fixed in `eb3f515`.
+
+Gate re-run: `npm test` 11/11; old skip conjunction 0 hits; `treeDiffers`
+3; `npm test` in CI 2. Human-verify item unchanged (pending).
--- a/docs/autopilot/2026-08-19-ax-136-receiver-byte-diff-report.md
+++ b/docs/autopilot/2026-08-19-ax-136-receiver-byte-diff-report.md
@@ -85,5 +85,34 @@ still copy a symlink — source is now listed (and its root `lstat`ed) first,
 two first-import tests added (`7ed6a67`); `test-receiver` gets an explicit
 `contents: read` token (`ec61f34`). Tests 13/13.
 
+## Iteration 2 — 2026-08-25 (local review)
+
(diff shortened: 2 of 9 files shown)


scripts/ctx-receive.mjs
```diff
@@ -144,16 +223,29 @@ function main() {
...
+  if (!opts.dryRun && opts.docsCommit) state[ORDERING_KEY] = opts.docsCommit;
+
+  const nextState = JSON.stringify(state, null, 2) + '\n';
+  const stateChanged =
+    !opts.dryRun && (!fs.existsSync(opts.state) || fs.readFileSync(opts.state, 'utf8') !== nextState);
+  if (stateChanged) fs.writeFileSync(opts.state, nextState);
 
   console.log(changed.length ? `\nChanged: ${changed.join(', ')}` : '\nNo changes.');
+  if (stateChanged && !changed.length) {
+    console.log(`State advanced to ${opts.docsCommit.slice(0, 12)} with no skill changes.`);
```
domitriusclark: Crashes when `--docs-commit` is omitted (it's optional per line 42): a fresh state file makes `stateChanged` true, and this line calls `.slice()` on `null`. Reproduced on this branch. The throw also lands before the `GITHUB_OUTPUT` write, so CI would see an empty `changed_count`.

Fix: guard on `opts.docsCommit`, and add a test that omits the flag — the helper currently always passes it.

scripts/ctx-receive.mjs
```diff
@@ -108,15 +177,17 @@ function main() {
...
     if (!sourceHash) fail(`${manifestPath}: missing generation.source_hash`);
 
-    const prev = state[grouping];
-    if (prev && prev.sourceHash === sourceHash && prev.importerVersion === importerVersion) {
-      console.log(`[skip] ${grouping}: unchanged (source_hash ${sourceHash.slice(0, 12)})`);
-      continue;
-    }
-
     const skillSrc = path.join(groupingDir, 'skill');
     if (!fs.existsSync(path.join(skillSrc, 'SKILL.md'))) {
```
domitriusclark, line 227: This check used to be behind the skip; now it runs for every grouping and `fail()` kills the whole run. One grouping with a manifest but no `skill/` folder yet blocks imports for all the others.

Fix: warn and `continue`, like the missing-manifest case above it.

scripts/ctx-receive.mjs
```diff
@@ -144,16 +223,29 @@ function main() {
...
+    !opts.dryRun && (!fs.existsSync(opts.state) || fs.readFileSync(opts.state, 'utf8') !== nextState);
+  if (stateChanged) fs.writeFileSync(opts.state, nextState);
 
   console.log(changed.length ? `\nChanged: ${changed.join(', ')}` : '\nNo changes.');
+  if (stateChanged && !changed.length) {
+    console.log(`State advanced to ${opts.docsCommit.slice(0, 12)} with no skill changes.`);
+  }
 
   if (process.env.GITHUB_OUTPUT) {
+    // `state_changed` exists so the workflow can commit an ordering-only
```
domitriusclark: Nothing reads `state_changed`, and the header comment points at a "monotonicity guard" in the workflow that doesn't exist. Also, line 230 assigns the commit unconditionally, so re-dispatching an older commit rewinds the key.

Fix: land the workflow change (and the guard) with this PR, or remove the output and key until they have a consumer.

scripts/ctx-receive.mjs
```diff
@@ -86,6 +110,44 @@ function unionAffects(changes) {
...
+// matter for the delta.
+function listFiles(dir) {
+  const out = [];
+  (function walk(current, prefix) {
+    for (const entry of fs.readdirSync(current, { withFileTypes: true })) {
+      const abs = path.join(current, entry.name);
+      const rel = prefix ? `${prefix}/${entry.name}` : entry.name;
+      if (entry.isDirectory()) {
+        walk(abs, rel);
+      } else if (entry.isFile()) {
```
domitriusclark, line 161: The comparison only sees regular files' bytes, but `cpSync` copies symlinks and permissions too. Since this is now the only import gate, a new/retargeted symlink or a chmod upstream is skipped forever — the same silent-skip class this PR fixes. And a byte diff in a file git normalizes or ignores gives `changed_count=1` with an empty stage, the empty-commit failure again.

Fix: fail loudly on symlinks (declare them unsupported), or copy unconditionally and let `git status --porcelain -- skills/` be the single judge of "changed".

scripts/ctx-receive.mjs
```diff
@@ -108,15 +177,17 @@ function main() {
...
-      fail(`${grouping}: changed but ${skillSrc}/SKILL.md is missing`);
+      fail(`${grouping}: ${skillSrc}/SKILL.md is missing`);
+    }
+
+    const prev = state[grouping];
+    const dest = path.join(opts.skillsDir, skill);
+
+    if (!treeDiffers(skillSrc, dest)) {
+      console.log(`[skip] ${grouping}: surface identical (source_hash ${sourceHash.slice(0, 12)})`);
+      continue;
```
domitriusclark, line 256: `importerVersion` is written but no longer compared here, so the README's "bump to force a re-import" knob is now a no-op.

Fix: make a bump count as changed, or remove the field and its README entry.

scripts/ctx-receive.test.mjs
```diff
@@ -0,0 +1,249 @@
...
+    assert.equal(state.lastImportedCommit, 'docs-commit-2');
+    assert.equal(
+      state[GROUPING].docsCommit,
+      'docs-commit-1',
+      'per-grouping provenance stays at the commit that actually imported it',
+    );
+  });
+
+  // Restore the ordering key so later assertions compare against a known state.
+  run(fixture);
```
domitriusclark: The subtests share one mutated fixture and only pass in this order — this bare `run()` exists just to undo the previous subtest, and its result is never asserted.

Fix: `buildFixture()` per subtest; it's already a pure factory.

Nit: the `test-receiver` job's path filters miss `package.json`, and a `setup-node` pin would make the Node 18+ floor real.

domitriusclark (commented):
Byte-diff is the right approach and the test suite is a great addition (6/6 passing locally). Six inline comments; two more that live in files this PR doesn't touch:

**`lastImportedCommit` is never persisted.** The workflow's commit step still fires only on `changed_count != '0'`, so an ordering-only advance is written on the runner and discarded — the exact case `state_changed` was added for. Either update the workflow condition in this PR or drop the ordering key until the consumer ships.

**`.ctx-gen/README.md` still says "unchanged hash → skip"** — the rule this PR deletes — and doesn't mention the new `lastImportedCommit` key. Worth updating in this PR.

seancdavis:
Thanks — all eight landed, plus CodeRabbit's `treeDiffers` one. Every finding checked out against the branch. Here's the map:

**Fixed**

- **`.slice()` on null `docsCommit`** — gone with the ordering key (below). Pinned by a new test that runs without `--docs-commit` and asserts the manifest-commit fallback. `ace7e5f`
- **Missing `SKILL.md` kills the run** — now `[skip] <grouping>: … SKILL.md is missing` + `continue`, same as the missing-manifest branch. Test: two-grouping fixture, B missing `SKILL.md`, A still imports. `a701fb9`
- **Symlinks / modes invisible to the gate** — took the "declare unsupported" option, not "git as judge": `listFiles()` now fails loudly on any non-regular entry in either tree, and `treeDiffers()` compares the executable bit (the only mode bit git tracks) alongside bytes. Tests for both. On the git-normalization point: `.gitattributes` has no `text=auto` and `.gitignore` has nothing under `skills/`, so byte-diff and git-diff agree here today; noted in the spec rather than guarded. `a701fb9`
- **`importerVersion` is a no-op knob** — removed rather than made-to-count: with a faithful byte copy no version bump can change output, and a forced re-import of identical bytes would only recreate the empty-commit failure. Dropped from the script, `config.json`, all 13 `state.json` entries, and the README. `a701fb9`, `1d9ae10`
- **Dest exists but isn't a directory** (CodeRabbit) — `treeDiffers()` uses `lstat` and returns true for a stray file or symlink at `skills/<name>`; the import's `rmSync`+`cpSync` replaces it. Test added. `a701fb9`
- **Order-dependent tests** — every case builds its own fixture and cleans up after itself; the bare `run()` is gone. `ace7e5f`, `eb3f515`
- **CI nits** — `package.json` added to the path filter; `actions/setup-node@…` (SHA-pinned, v7.0.0) with `node-version: 18` so the Node 18 floor is actually exercised. `e22e3b4`
- **`.ctx-gen/README.md`** — `state.json` is now described as a provenance log; the delta is the byte comparison. `1d9ae10`

**Moved, not fixed here**

- **`lastImportedCommit` / `state_changed` with no consumer** — you're right that #115 alone had a dead output and an unguarded rewind. The consumer is #110 (`ctx-receiver-monotonic`: the workflow guard + `ctx-ordering-guard.mjs`), so I reverted the writer out of this PR (`1a0b93e`) and it moves into #110, which becomes self-contained (writer + guard + workflow commit condition). This PR is now purely the byte-diff gate. #110 will need a rebase once this merges.

**Not changed, on purpose**

- CodeRabbit's docstring-coverage warning — the functions carry explanatory comments in the file's existing style; no JSDoc added.

Spec amended (`docs/autopilot/…-byte-diff.md`, Program design) to record the symlink/exec-bit policy, the `importerVersion` removal, and the warn-not-fail on missing `SKILL.md`; run report has an iteration section. Tests: 11/11. Codex security pass clean; simplicity pass had one low finding (test-helper state), taken.

scripts/ctx-receive.mjs
seancdavis: Fixed. The ordering key and this log line moved to #110 (reverted here in `1a0b93e`; the port in #110 `16be109` carries the `&& opts.docsCommit` guard). A test that omits `--docs-commit` is in `ace7e5f`.

scripts/ctx-receive.mjs
seancdavis: Fixed in `a701fb9`: `[skip] <grouping>: … SKILL.md is missing` then `continue`, same as the manifest branch. Two-grouping test covers A importing while B is skipped.

scripts/ctx-receive.mjs
seancdavis: Agreed. The consumer is #110 (`ctx-ordering-guard.mjs` + the workflow step), so the writer moved there instead of landing the workflow here: reverted in `1a0b93e`, ported to #110 in `16be109`. This PR is byte-diff only now.

scripts/ctx-receive.mjs
seancdavis: Took the first option in `a701fb9`: `listFiles()` fails on any non-regular entry, and `treeDiffers()` compares the executable bit (the one mode bit git tracks) alongside bytes. On normalization: `.gitattributes` has no `text=auto` and `.gitignore` has nothing under `skills/`, so byte-diff and git-diff agree today — recorded in the spec rather than guarded. CodeRabbit then caught that the first-import short-circuit runs before the source check; fixing that in a follow-up commit.

scripts/ctx-receive.mjs
seancdavis: Removed in `a701fb9` + `1d9ae10` — script, `config.json`, all 13 `state.json` entries, README. Making a bump count as changed would re-import identical bytes and hit the empty-commit failure, so the knob goes rather than the docs.

scripts/ctx-receive.test.mjs
seancdavis: Fixed in `ace7e5f` (+ `eb3f515`): every case builds its own fixture and cleans up; the bare `run()` is gone. Nit taken in `e22e3b4`: `package.json` in the path filter, `setup-node` SHA-pinned at Node 18.

docs/autopilot/2026-08-19-ax-136-receiver-byte-diff-report.md
```diff
@@ -0,0 +1,53 @@
...
+## Completeness gate
+
+- `npm test` → 5/5 pass (orchestrator-run).
+- `rg "prev.sourceHash === sourceHash && prev.importerVersion"` → 0 hits.
+- `rg -c "treeDiffers"` → 3 (≥ 2).
+- `rg -c "npm test" .github/workflows/validate-skills.yml` → 2 (≥ 1).
+- human-verify (for Sean): after netlify/docs#801 merges, the next receive
+  run's rolling sync PR must contain the Next.js fetch-skew qualification.
+
+No unmet items; no deferred findings beyond the rejected CI-job fold above.
```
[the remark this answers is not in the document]
seancdavis: Marked pending in `142024e`; the closing line now says the human-verify item is the one open item.

scripts/ctx-receive.mjs
```diff
@@ -86,11 +101,62 @@ function unionAffects(changes) {
...
+// files, or any shared-path file with different bytes or a different
+// executable bit. A missing `destDir` (e.g. first import), or one that is
+// not a directory (a stray file or symlink at `skills/<name>`), counts as
+// different — the caller's rmSync + cpSync then replaces it.
+function treeDiffers(srcDir, destDir) {
+  const destStat = fs.lstatSync(destDir, { throwIfNoEntry: false });
+  if (!destStat?.isDirectory()) return true;
+
+  const srcFiles = listFiles(srcDir);
+  const destFiles = listFiles(destDir);
```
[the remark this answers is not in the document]
seancdavis: Confirmed and fixed in `7ed6a67`: `treeDiffers()` lists (and so validates) the source before the destination short-circuit, and `listFiles()` `lstat`s its root so a symlinked `skill/` fails too. Two first-import tests added — nested symlink, symlinked root — both assert the destination is never created.

.github/workflows/validate-skills.yml
```diff
@@ -56,3 +59,20 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10
+        with:
+          persist-credentials: false
+
+      # 18 is the floor the script header promises; running on it keeps that claim honest.
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 18
```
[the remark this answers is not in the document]
seancdavis: Added in `ec61f34` — `permissions: contents: read` on `test-receiver` only; the pre-existing `validate` job is out of this PR's scope.

seancdavis:
Second round, from a local review of the branch:

- **Previously imported grouping loses `skill/SKILL.md`** — the warn-and-skip from round one was too broad: it also covered a grouping that had already been imported, so an upstream deletion left the old skill in place on a green run. Now the skip applies only to a never-imported grouping (no state entry, no `skills/<name>`); otherwise the run fails. Two tests. `55cfe8a`
- **AX-136 covers `context.md`/`system.md`, not only `skill/**`** — a hand edit to the intermediates with no regenerated skill was still a silent no-op. Decision (Sean): this repo must not fail on no-ops, so the receiver now hashes the two intermediates into `state.json` (`intermediateHash`) and, when that hash moves while `skill/` is byte-identical, prints `[warn] …` (plus a `::warning::` annotation on Actions) once per upstream change and imports nothing. Legacy state entries are seeded silently. The receiver can't tell "forgot to regenerate" from "regeneration was a no-op", which is why this is a warning; a hard check belongs in docs CI. `6643f68`, `11b1738`, `b74cec5` (both intermediates deleted counts as a move, too)
- **#110's description** said the ordering key was introduced here — rewritten to say the writer lives in #110.

Known limitation until #110 lands: the workflow commits `state.json` only when `changed_count != 0`, so a hash-only update is discarded on the runner and the warning repeats each run until the next import; #110's `state_changed` gate makes it fire once.

Tests: 22/22. Codex security clean; simplicity flagged the hash→null gap (taken) and the persistence-before-#110 point (kept as-is, documented in the code comment).

domitriusclark (approved):
All eight review findings verified fixed on the latest push — reran the suite locally (22/22). The revert of the ordering key out of this PR is the right split; will pick up that contract in the #110 review. The intermediate-drift warning got its own adversarial pass: separator-safe hashing, null treated as a real value, dry-run doesn't persist — clean. The one repeat-until-#110 limitation is documented inline. Nice iteration.

PR #117, opened 2026-08-27
fix(ctx-pipeline): surface guard skips, fail closed on divergence, validate --docs-commit

domitriusclark:
Two small follow-ups from the PR #110 review (monotonicity guard for the Stage 2 receiver).

## AX-159 — observability in the guard skip path

A stale-delivery skip in `scripts/ctx-ordering-guard.mjs` was a plain stdout line inside a green run, while the less-dangerous `skip_guard` bypass already got a `::warning`. Now:

- A skipped delivery emits a `::warning title=ctx-ordering-guard::` annotation and a `GITHUB_STEP_SUMMARY` entry, so a pattern of skips is visible from the run list.
- The non-descent case splits two ways. **Stale** (incoming is an *ancestor* of the recorded position): a newer dispatch already imported past it — skip, self-healing. **Diverged** (neither descends from the other — the shape @seancdavis described, where a netlify/docs history rewrite leaves `lastImportedCommit` resolvable but main never descends from it again): previously this skipped green on every future delivery forever; it now **fails closed** with a message pointing at the `skip_guard` manual-recovery path, which resets the ordering baseline on import.

## Full-SHA validation of `--docs-commit`

`scripts/ctx-receive.mjs` wrote `state[lastImportedCommit] = opts.docsCommit` unvalidated, so a manual local run with a short SHA wrote a malformed ordering key and the next CI run failed closed. `--docs-commit` is now rejected up front unless it matches `/^[0-9a-f]{40}$/` (seancdavis's inline nit on PR #110, deferred to avoid conflicting with #115, now merged).

## Tests

Both zero-dep suites extended (`npm test`, 46 passing):
- guard: stale skip message + `::warning`/step-summary assertions (on and off Actions), proceed/fail paths write no summary, unrelated-history case now asserts fail-closed with `skip_guard` guidance.
- receiver: short SHA, non-hex, uppercase, and branch-name `--docs-commit` values are rejected before anything is written; existing fixtures moved to full hex SHAs.

Ref: AX-97, AX-159.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ordering checks so stale deliveries now surface a warning and summary note, while diverged histories fail safely instead of passing silently.
  * Tightened validation for commit references to reject invalid values before any changes are applied.
* **Documentation**
  * Clarified the handling of skipped, diverged, and unrecoverable ordering cases in workflow notes.
* **Tests**
  * Added coverage for warning behavior, fail-closed behavior, and invalid commit reference validation.

### DIFF
--- a/scripts/ctx-ordering-guard.mjs
+++ b/scripts/ctx-ordering-guard.mjs
@@ -7,12 +7,20 @@
 //
 //   proceed  incoming descends from the last imported commit (or nothing has
 //            been imported yet — bootstrap)
-//   skip     incoming does not descend from the last imported commit
-//            (stale or unrelated delivery); exit 0 with skip=1
+//   skip     incoming is an ANCESTOR of the last imported commit (a stale
+//            delivery — a newer dispatch already imported past it). Exit 0
+//            with skip=1; self-healing, since the next on-line delivery
+//            proceeds. On Actions the skip is surfaced as a ::warning
+//            annotation and a GITHUB_STEP_SUMMARY entry (AX-159), not just a
+//            green log line.
 //   fail     the recorded position exists but cannot be validated — malformed
 //            JSON, a malformed ordering key, or a recorded commit the docs
-//            checkout cannot resolve. Fail-closed: silence here is how
-//            rollbacks happen. Exit 1.
+//            checkout cannot resolve. ALSO when incoming and the recorded
+//            position have DIVERGED (neither descends from the other, e.g.
+//            after a docs history rewrite where the old commit still
+//            resolves): no future delivery on that line can ever descend from
+//            the recorded position, so skipping green would recur forever.
+//            Fail-closed: silence here is how rollbacks happen. Exit 1.
 //
 // Ordering comes from the top-level `lastImportedCommit` key in state.json —
 // the ordering AUTHORITY written by ctx-receive.mjs on every run. Per-grouping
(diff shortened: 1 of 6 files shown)


scripts/ctx-ordering-guard.mjs
```diff
@@ -118,7 +141,22 @@ function main() {
...
+    // rewrite where the recorded commit still resolves, or a foreign branch)
+    // never self-heals: every future delivery on that line would skip green
+    // forever, so it fails closed instead.
+    const reverse = git(opts.docs, ['merge-base', '--is-ancestor', opts.incoming, last]);
+    if (reverse === 0) {
+      const msg = `incoming ${opts.incoming} is an ancestor of last imported ${last} — stale delivery, skipping`;
+      surfaceSkip(msg);
+      emit(true, msg);
+    } else if (reverse === 1) {
+      fail(`incoming ${opts.incoming} and last imported ${last} have diverged — no delivery on this line can ever descend from the recorded position (docs history rewrite?); failing closed. If the divergence is expected, re-run manually with skip_guard to reset the ordering baseline`);
```
seancdavis, line 156: **Consider** — the recovery pointer only lands on this branch. The `cat-file` fail above is the one a real docs history rewrite is more likely to hit, since a rewrite usually orphans the old commit so it never resolves at all, and that message says nothing about `skip_guard`. Same operator, same recovery, one of the two paths tells them.

seancdavis (approved):
One thing outside this PR: when the receive workflow goes red, nothing says so. Slack only watches the docs side, so the docs run reports success while the import here quietly stops — filed as AX-162. Not something this PR needs to carry.

PR #120, opened 2026-09-10
feat(ctx-pipeline): post receive outcomes to Slack (EX-3057)

domitriusclark:
Follow-on to #117 for EX-3057. docs' dispatch to us is fire-and-forget, so the docs-side notifier (AX-138) says "delivered" the moment the dispatch is accepted. If the receive run here fails, nothing says so: `workflow_run` can't watch across repos, the run has no PR or watched branch, and the failure mail goes to the dispatch token's identity. With #117's fail-closed divergence branch that's now a state you can sit in for weeks - every dispatch after a docs history rewrite fails the same way and Slack keeps reporting success.

## What changed

- `.github/workflows/ctx-pipeline-notify.yml` - `workflow_run` watcher on "Receive agent-context". Posts every outcome to `#notify-context-pipeline`; the only silent case is a run that concluded `skipped` (dispatch with `CTX_PIPELINE` off). `workflow_dispatch` replays any run ID, dry-run by default.
- `scripts/ctx-notify.mjs` - the classifier. Five shapes: 📥 IMPORTED, 💤 NO-OP, ⏭️ SKIPPED (stale delivery), 🔴 FAILED, ⚠️ UNCLASSIFIED. Failures name the step and what to do about it; the guard's fail-closed branch calls out `skip_guard`, since that state recurs on every dispatch until a human resets it. One field per line: status, docs sha + trigger, detail, run URL, PR URL when there is one. The docs sha pairs the message with the docs-side DELIVERED line.
- `.github/workflows/ctx-pipeline-receive.yml` - the PR step now outputs its URL, and two always-run, `continue-on-error` steps upload a `ctx-receive-outcome` artifact (docs sha, groupings, PR URL). Classification reads job step conclusions from the Actions API, so runs that die at checkout still classify; the artifact only enriches the message.
- `scripts/ctx-notify.test.mjs` - every shape, every known failing step, degradation without the artifact, and the exact message lines. Wired into `npm test` and the validate-skills path filter.

## Plain text on purpose

The channel's webhook is a Slack Workflow Builder trigger, not a classic incoming webhook (EX-3065). It drops `text` into a template as plain text, which is why every docs-side line in the channel today renders its links as `<<url>|run 123>`. A probe post confirmed the rest: newlines break lines, bare URLs auto-link, and `*bold*` / `&amp;` / `<url|label>` all come through literally. So this message uses bare URLs and newlines and nothing else. The one payload-derived field (`docs_ref`) has angle brackets stripped rather than entity-escaped. The docs-side notifier has the same rendering problem; that's a separate fix in `netlify/docs`.

## Why not reuse docs' notify.mjs

The ticket asks. Its classifier is entirely docs-specific (gate evidence artifact, case-B PR lookup); only the watcher pattern carries over, and after the plain-text finding its formatting layer doesn't either. Pulling private docs code into a public repo's CI just to post one message didn't seem worth it.

## Verified

- `npm test`: 68 pass, 23 new.
- Dry-run replay against real runs: 32073913019 (green dispatch) → IMPORTED; 31633992205 and 31633016533 (red manual) → FAILED at "Checkout context-and-tools".
- The outcome step's inline `node -e` snippet run locally with sample env and fed through the formatter end to end.
- One probe message with the exact IMPORTED layout posted through the real webhook by hand (tagged `[test EX-3057 · ignore]` in the channel).

## Before this does anything

- `SLACK_WEBHOOK_URL` needs to exist in a `ctx-pipeline` environment on this repo (same webhook docs uses; GitHub creates the environment on the first run that references it). When creating it, restrict deployment branches to `main` - this is a public repo and the environment is what keeps the secret off every other branch. Until the secret exists the notifier prints a dry-run message and exits 0.
- Neither trigger fires until the file is on `main`. First live check: `workflow_dispatch` with any run ID and `dry_run` on.
- Further formatting (bold, link labels) is a Workflow Builder template change under EX-3065, not something this payload can do.

Ref: EX-3057, AX-138, AX-159.

## Summary by CodeRabbit

* **New Features**
  * Slack notifications now include unclassified outcomes.
  * Notifications support manual replay and dry runs.
  * Receive results remain available as downloadable artifacts for 30 days.
  * Notifications stay inactive when Slack configuration is unavailable.

* **Bug Fixes**
  * Restricted notifications to trusted workflow runs from the current repository.
  * Prevented stale artifacts from being used for reruns.
  * Added timeouts and retries for GitHub and Slack requests.

* **Tests**
  * Expanded validation for workflow changes, trust boundaries, reruns, and outcome classification.

### DIFF
--- a/.ctx-gen/README.md
+++ b/.ctx-gen/README.md
@@ -34,6 +34,13 @@ AXIS scenarios.)
   draft PR when something changed. That PR runs the existing `validate-skills`
   and `build-generated-outputs` (cursor/codex parity) gates; a human reviews
   and merges. Rollback = revert.
+- **`../.github/workflows/ctx-pipeline-notify.yml`** — a `workflow_run`
+  watcher that posts every receiver outcome (imported / no-op / stale skip /
+  failed) to `#notify-context-pipeline`, the same channel the docs side
+  reports to. docs' dispatch is fire-and-forget, so without this a failed
+  import was invisible while Slack kept saying "delivered" (EX-3057). Logic in
+  `../scripts/ctx-notify.mjs`; inert until `SLACK_WEBHOOK_URL` exists in the
+  `ctx-pipeline` environment.
 
 ## Triggers
 
--- a/.ctx-gen/README.md
+++ b/.ctx-gen/README.md
@@ -36,7 +36,7 @@ AXIS scenarios.)
   and merges. Rollback = revert.
 - **`../.github/workflows/ctx-pipeline-notify.yml`** — a `workflow_run`
   watcher that posts every receiver outcome (imported / no-op / stale skip /
-  failed) to `#notify-context-pipeline`, the same channel the docs side
+  failed / unclassified) to `#notify-context-pipeline`, the same channel the docs side
   reports to. docs' dispatch is fire-and-forget, so without this a failed
   import was invisible while Slack kept saying "delivered" (EX-3057). Logic in
   `../scripts/ctx-notify.mjs`; inert until `SLACK_WEBHOOK_URL` exists in the
--- a/.github/workflows/ctx-pipeline-notify.yml
+++ b/.github/workflows/ctx-pipeline-notify.yml
@@ -45,10 +45,22 @@ permissions:
 
 jobs:
   notify:
-    # A dispatch that arrived with CTX_PIPELINE off skips the receive job and
-    # the run concludes "skipped" — nothing to report, so don't spin up a
-    # runner to decide silence. Every other outcome is a message shape.
-    if: github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion != 'skipped'
+    # Trust boundary first: workflow_run matches by workflow NAME, so a fork PR
+    # that adds a pull_request trigger to the receive workflow (or ships a
+    # same-named file) would complete a run this watcher fires on — with the
+    # webhook secret, reading the fork's step names and artifact. Only runs
+    # from this repository on the receive workflow's two legitimate triggers
+    # get a runner; ctx-notify.mjs re-checks the same rule before posting.
+    # Then: a dispatch that arrived with CTX_PIPELINE off skips the receive job
+    # and the run concludes "skipped" — nothing to report, so no runner for
+    # that either. Every other outcome is a message shape.
+    if: >
+      github.event_name == 'workflow_dispatch' ||
+      (
+        github.event.workflow_run.head_repository.full_name == github.repository &&
+        contains(fromJSON('["repository_dispatch","workflow_dispatch"]'), github.event.workflow_run.event) &&
+        github.event.workflow_run.conclusion != 'skipped'
+      )
     runs-on: ubuntu-latest
     environment: ctx-pipeline
     steps:
--- a/.github/workflows/ctx-pipeline-notify.yml
+++ b/.github/workflows/ctx-pipeline-notify.yml
@@ -1,10 +1,8 @@
 name: ctx-pipeline notify
 
-# EX-3057: docs dispatches to us fire-and-forget, so the docs-side notifier
-# (AX-138) posts "delivered" the moment the dispatch is accepted — a receive
-# failure here was invisible while Slack kept reporting success. This watcher
-# posts every receive outcome to the same channel, #notify-context-pipeline,
-# so both halves of the pipeline report in one place.
+# EX-3057: post every "Receive agent-context" outcome to
+# #notify-context-pipeline, the channel the docs-side notifier (AX-138)
+# already reports to. Why it exists: see the header of scripts/ctx-notify.mjs.
 #
 # A separate workflow_run watcher, not notify steps inside the receiver: it
 # fires on every terminal state — including runs that die at checkout,
--- a/.github/workflows/ctx-pipeline-notify.yml
+++ b/.github/workflows/ctx-pipeline-notify.yml
@@ -1,0 +1,69 @@
+name: ctx-pipeline notify
+
+# EX-3057: docs dispatches to us fire-and-forget, so the docs-side notifier
+# (AX-138) posts "delivered" the moment the dispatch is accepted — a receive
+# failure here was invisible while Slack kept reporting success. This watcher
+# posts every receive outcome to the same channel, #notify-context-pipeline,
+# so both halves of the pipeline report in one place.
+#
+# A separate workflow_run watcher, not notify steps inside the receiver: it
+# fires on every terminal state — including runs that die at checkout,
+# cancels, and timeouts — and a bug here can never redden the receiver
+# itself. All logic lives in scripts/ctx-notify.mjs (thin-caller rule); the
+# five message shapes are documented there and tested by
+# scripts/ctx-notify.test.mjs (runs in validate-skills).
+#
+# workflow_run and workflow_dispatch both run this file's copy on the default
+# branch, so neither trigger works before merge. To test pre-merge, replay a
+# historical run locally:
+#   node scripts/ctx-notify.mjs --run-id <id> --repo netlify/context-and-tools --dry-run
+# After merge, workflow_dispatch replays any run ID — dry_run prints the exact
+# message instead of posting.
+#
+# Inert until SLACK_WEBHOOK_URL exists in the ctx-pipeline environment. It is
+# the same Slack webhook the docs side posts through (docs keeps its copy in
+# its ctx-gen environment); environment scoping means the secret only injects
+# into jobs that declare the environment, which matters in a public repo.
+# Without the secret, ctx-notify.mjs prints the message as a dry run and
+# exits 0.
+
+on:
+  workflow_run:
+    workflows: ["Receive agent-context"]
+    types: [completed]
+  workflow_dispatch:
+    inputs:
+      run_id:
+        description: "Run ID of a Receive agent-context run to classify (replay)"
+        required: true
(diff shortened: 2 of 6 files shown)


.github/workflows/ctx-pipeline-notify.yml
```diff
@@ -0,0 +1,67 @@
...
+permissions:
+  contents: read
+  actions: read # run metadata + the ctx-receive-outcome artifact
+
+jobs:
+  notify:
+    # A dispatch that arrived with CTX_PIPELINE off skips the receive job and
+    # the run concludes "skipped" — nothing to report, so don't spin up a
+    # runner to decide silence. Every other outcome is a message shape.
+    if: github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion != 'skipped'
```
seancdavis: [blocking] Nothing checks where the watched run came from, so a fork PR can post through this webhook.

A fork PR can add `pull_request` to the receive workflow's triggers ([GitHub Security Lab](https://securitylab.github.com/resources/github-actions-new-patterns-and-mitigations/) covers PRs rewriting a triggering workflow). When that run finishes, this job fires on `main` with the secret and reads the PR's artifact and step names, so the `main`-only environment doesn't help here. `changed` and the failed step name go into the message as-is and bare URLs auto-link, which is enough for a fake pipeline alert with an outsider's link in it. The same watcher is fine in docs because that repo is private.

scripts/ctx-notify.test.mjs
```diff
@@ -0,0 +1,285 @@
...
+
+import {
+  RECEIVE_WORKFLOW,
+  classifyRun,
+  formatMessage,
+  parseOutcome,
+  stripMarkup,
+} from './ctx-notify.mjs';
+
+// Step names exactly as .github/workflows/ctx-pipeline-receive.yml declares
```
seancdavis: [consider] This list is copied by hand, so nothing fails if a step in `ctx-pipeline-receive.yml` gets renamed, and validate-skills doesn't run on that file anyway.

A rename of the PR step turns green runs into ⚠️ UNCLASSIFIED, and a rename of the guard step drops the skip_guard line from the one failure where it matters.

scripts/ctx-notify.mjs
```diff
@@ -0,0 +1,277 @@
...
+//   ⏭️ SKIPPED       stale delivery — the monotonicity guard refused an older
+//                    docs commit (AX-159); self-heals on the next dispatch
+//   🔴 FAILED        with the failing step and, where the step is known, what
+//                    to do about it (the guard's fail-closed branch calls out
+//                    the skip_guard recovery, because that state recurs on
+//                    every later dispatch until a human resets it)
+//   ⚠️ UNCLASSIFIED  cancelled / timed out / unrecognized job layout
+// Runs with conclusion "skipped" (CTX_PIPELINE off) post nothing.
+//
+// Message layout, one field per line (the channel is scraped as well as
```
seancdavis: [consider] What scrapes this channel? The order is called a contract here, but it doesn't match the docs-side lines (one `·`-joined line, run link second).

scripts/ctx-notify.mjs
```diff
@@ -0,0 +1,277 @@
...
+    if (!webhook) console.log('SLACK_WEBHOOK_URL unset — inert until the secret exists in the ctx-pipeline environment');
+    console.log(`notify (dry-run): ${message}`);
+    return;
+  }
+  const res = await fetch(webhook, {
+    method: 'POST',
+    headers: { 'content-type': 'application/json' },
+    body: JSON.stringify({ text: message }),
+  });
+  if (!res.ok) {
```
seancdavis: [consider] One failed Slack POST or `gh api` call means no message at all, and the red notify run is as easy to miss as the receive failures this PR is fixing.

scripts/ctx-notify.mjs
```diff
@@ -0,0 +1,277 @@
...
+  if (name.startsWith(STEP.checkoutDocs)) {
+    // docs_ref echoes the dispatch payload — untrusted, so it must not carry
+    // Slack markup (<!channel>) into the message.
+    const ref = outcome?.docs_ref ? stripMarkup(truncate(outcome.docs_ref, 60)) : 'the requested ref';
+    return `could not check out netlify/docs at ${ref} — DOCS_READ_TOKEN expired, or the ref no longer exists (docs history rewrite?)`;
+  }
+  if (name.startsWith(STEP.guard))
+    return 'monotonicity guard failed closed — docs history diverged from lastImportedCommit, or state.json is unreadable. Every later dispatch fails the same way until a manual run with skip_guard resets the baseline';
+  if (name.startsWith(STEP.import))
+    return 'import failed — a previously imported grouping vanished upstream, or an unsupported entry (symlink) in the skill tree; see run';
```
seancdavis: [nit] ctx-receive fails in more ways than these two. A skill `name` that doesn't match the mapping, or a missing `generation.source_hash`, would get reported as a vanished grouping or a symlink.

seancdavis (changes requested):
One blocking comment and a few suggestions.

.github/workflows/ctx-pipeline-notify.yml
domitriusclark: Good catch - I'd only thought about who could read the secret, not who could produce a run the watcher fires on. Fixed in 1cd6e8d at two layers: the job `if:` now requires `head_repository == github.repository` and the run event to be `repository_dispatch` or `workflow_dispatch`, and `ctx-notify.mjs` re-checks the same rule before reading steps or the artifact (covers manual replays too). Added a tripwire test that the receive workflow's trigger set equals those two, so adding `pull_request` to it later fails the suite.

scripts/ctx-notify.test.mjs
domitriusclark: Yeah that was lazy. The test now parses `- name:` lines out of the receive workflow and asserts each prefix matches exactly one step, and both workflow files are in the validate-skills path filter. Rename the guard step and CI goes red.

scripts/ctx-notify.mjs
domitriusclark: Nothing, AFAIK. I lifted "contract / scraped" from the docs-side header without checking. Dropped the claim - the order stays fixed so you can pair a receive message with the docs line, and that's the only reason.

scripts/ctx-notify.mjs
domitriusclark: Partly fixable. Added retries (3x, backoff) and a 10s timeout on the Slack POST and the gh api reads, which covers the transient stuff. If Slack is actually down we can't report that through Slack, so the red notify run is the floor - same as the docs side. Said so in the header rather than pretend otherwise. Open to a fallback (issue? email?) if you think it's worth a follow-up.

scripts/ctx-notify.mjs
domitriusclark: Fair. It just says "ctx-receive exited non-zero, see run" now instead of guessing.

.ctx-gen/README.md
```diff
@@ -34,6 +34,13 @@ AXIS scenarios.)
   draft PR when something changed. That PR runs the existing `validate-skills`
   and `build-generated-outputs` (cursor/codex parity) gates; a human reviews
   and merges. Rollback = revert.
+- **`../.github/workflows/ctx-pipeline-notify.yml`** — a `workflow_run`
+  watcher that posts every receiver outcome (imported / no-op / stale skip /
```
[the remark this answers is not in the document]
domitriusclark: Done in 1cd6e8d.

scripts/ctx-notify.mjs
```diff
@@ -0,0 +1,274 @@
...
+    console.error('usage: ctx-notify.mjs --run-id <id> [--repo owner/name] [--dry-run]');
+    process.exit(2);
+  }
+  if (!/^\d+$/.test(args.runId)) {
+    console.error(`--run-id must be numeric, got ${JSON.stringify(args.runId)}`);
+    process.exit(2);
+  }
+  const run = ghJson(['api', `repos/${repo}/actions/runs/${args.runId}`]);
+  const { jobs } = ghJson(['api', `repos/${repo}/actions/runs/${args.runId}/jobs?per_page=100`]);
+  const outcome = run.conclusion === 'skipped' ? null : loadOutcome(repo, args.runId);
```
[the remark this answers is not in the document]
domitriusclark: Done in 1cd6e8d - outcome.json carries `run_attempt` and the notifier ignores it on mismatch.

scripts/ctx-notify.mjs
```diff
@@ -0,0 +1,274 @@
...
+  if (args.dryRun || process.env.DRY_RUN === 'true' || !webhook) {
+    if (!webhook) console.log('SLACK_WEBHOOK_URL unset — inert until the secret exists in the ctx-pipeline environment');
+    console.log(`notify (dry-run): ${message}`);
+    return;
+  }
+  const res = await fetch(webhook, {
+    method: 'POST',
+    headers: { 'content-type': 'application/json' },
+    body: JSON.stringify({ text: message }),
+  });
```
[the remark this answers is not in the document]
domitriusclark: Done in 1cd6e8d - 10s timeout plus retries on the POST and the gh api reads.

scripts/ctx-notify.mjs
seancdavis: I don't think it's worth a follow-up. The retries cover the blips, and if Slack itself is down we'll hear about it some other way.

seancdavis (approved):
🤘

